PR #12, opened 2015-03-11
Fix oddly camel cased GitHub URL

skalnik:
Noticed this while reading the docs, and seems to the only link formatted this way.


thedaniel:
:+1:

PR #15, opened 2015-03-12
Fix activationEvents -> activationCommands

olmokramer:
Change `activationEvents` to `activationCommands` and update the description to match current behavior.


nathansobo:
Thanks very much!

PR #21, opened 2015-03-15
docs(windows) atom on windows is just as awesome as the mac version

basarat:
No more Chocolatey. Also it auto updates :heart:


nathansobo:
Thanks @basarat! @schacon This seems like it could warrant a new docs build since it's a fairly big correction.

schacon:
Sounds good.

schacon:
This is deployed

PR #22, opened 2015-03-15
`ctrl-cmd` makes little sense. Split `ctrl` vs. `cmd`

basarat:
I recommend using just `cmd` everywhere. That seems to be what the rest of the manual is following. But if you want to mention both then perhaps this is the way to go. Let me know in either case :heart:

### DIFF
--- a/book/02-using-atom/sections/02-text.asc
+++ b/book/02-using-atom/sections/02-text.asc
@@ -90,7 +90,7 @@ There are a handful of cool keybindings for basic text manipulation that might c
 
 `cmd-J`:: Join the next line to the end of the current line
 
-`ctrl-cmd-up`, `ctrl-cmd-down`:: Move the current line up or down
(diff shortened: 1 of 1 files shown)


izuzak:
There actually is a `ctrl-cmd-up` keybinding and it moves the current line up by one line. See:

https://github.com/atom/atom/blob/587ebd753840517dca899982d6a4f3931a1f1b39/keymaps/darwin.cson#L152-L153

If you hit just `cmd-up`, then the cursor will go to the first character of the first line.

https://github.com/atom/atom/blob/587ebd753840517dca899982d6a4f3931a1f1b39/keymaps/darwin.cson#L65-L66

Does that explain things?

basarat:
> Does that explain things?

Thanks. That clears things up. On windows its defined as `ctrl-up` which confused me. https://github.com/atom/atom/blob/587ebd753840517dca899982d6a4f3931a1f1b39/keymaps/win32.cson#L112-L113

PR #25, opened 2015-03-18
invalid href to `atom/settings-view` at `01-packages-themes.asc`

dasilvacontin:
The parenthesis was being included in the link.

### DIFF
--- a/book/02-using-atom/sections/01-packages-themes.asc
+++ b/book/02-using-atom/sections/01-packages-themes.asc
@@ -1,7 +1,7 @@
 [[_atom_packages]]
 === Atom Packages
 
-First we'll start with the Atom package system. As we mentioned previously, Atom itself is a very basic core of functionality that ships with a number of useful packages that add new features like the Tree view and the Settings view (which, by the way, can be found here: https://github.com/atom/settings-view).
+First we'll start with the Atom package system. As we mentioned previously, Atom itself is a very basic core of functionality that ships with a number of useful packages that add new features like the https://github.com/atom/tree-view[Tree View] and the https://github.com/atom/settings-view[Settings View].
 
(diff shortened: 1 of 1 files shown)


book/02-using-atom/sections/01-packages-themes.asc
```diff
@@ -1,7 +1,7 @@
 [[_atom_packages]]
 === Atom Packages
 
-First we'll start with the Atom package system. As we mentioned previously, Atom itself is a very basic core of functionality that ships with a number of useful packages that add new features like the Tree view and the Settings view (which, by the way, can be found here: https://github.com/atom/settings-view).
+First we'll start with the Atom package system. As we mentioned previously, Atom itself is a very basic core of functionality that ships with a number of useful packages that add new features like the Tree view and the Settings view (which, by the way, can be found here: https://github.com/atom/settings-view[https://github.com/atom/settings-view]).
```
izuzak: What do you think about using `Settings view` for the link, and deleting the rest? So, something like: 

>  ... add new features like the https://github.com/atom/tree-view[Tree view] and the https://github.com/atom/settings-view[Settings view].

dasilvacontin: I was just fixing the href, but I agree that `url[title]` is simpler and easier to read. I'll make the change now.

dasilvacontin: I think it makes sense to use the CapWords convention for the package title, like `url[Tree View]` or `url[Settings View]`, since they also use that convention on their respective repo READMEs, e.g. in https://github.com/atom/settings-view.

Do you agree with that?

izuzak: > I think it makes sense to use the CapWords convention for the package title,

Yep, agreed. :+1: Thanks for caring about the details. :heart:

dasilvacontin: Well, I'll just push the change with CapWords and hope for a thumbs up.

dasilvacontin: Oh. Didn't see the reply. "Keep it logically awesome" :)

PR #37, opened 2015-03-25
Add section on Services API

jlord:
Adding the section on the Services API from the old docs. But I don't know what I'm doing :monkey_face: I've tried running it locally and checking the PDF but I still don't see my addition. Halp. 

cc @schacon @kevinsawicki


schacon:
Awesome. You have to add new sections to be included in the main chapter files so they're pulled in when compiled, so I added that for you. Should be good now.

jlord:
Awesome, thanks @schacon! 

I just moved it to a different section in the book (Behind Atom) and am gonna publish and ship the blog post on Services API!

PR #49, opened 2015-04-18
Update 02-text.asc

magicae:
Fix typo


mnquintana:
Thanks @magicae! :ship:

PR #58, opened 2015-05-06
Update references to atom-shell to Electron

lee-dohm:
I'm not sure if the `outline.txt` is the plan-of-record anymore, but I figured this should be updated if it was still being used to record ideas.


kevinsawicki:
:+1:

PR #61, opened 2015-05-22
Add a section about capturing CPU profiles

nathansobo:
This PR includes a pretty big animated gif, but the description is a lot more compelling with a visual aid. The first frame would be useful in static settings. Do people that know more about the docs stack want to chime in on whether this is a good idea or not?


lee-dohm:
Tried it out in the following formats:

* ePub (as viewed in iBooks) &mdash; GIF showed first frame until double-clicked upon, then it shows the full animation
* Mobi (Kindle for Mac) &mdash; Kindle for Mac crashed (need to investigate if this is caused by the GIF or not)
* PDF (Mac Preview app) &mdash; GIF does not show at all

lee-dohm:
Mobi version crashes Kindle for Mac even on `master`, so it doesn't look like it is caused by this change.

izuzak:
@nathansobo I'd :heart: to see this in the docs so that we can link to it in issues. Thanks for adding this! :zap: 

In order to work around the problem with the GIF in PDF output (great catch, @lee-dohm!), what do you think about:

1. adding a screenshot instead of a GIF. I think I'd go with a screenshot which shows the Profiles tab right before you click "Start".
2. linking to Chrome's docs on [CPU profiling](https://developer.chrome.com/devtools/docs/cpu-profiling). Linking to those docs would be useful both to user who want to make sure they're doing things correctly and to user who want to dig deeper into this stuff.

:thought_balloon:

nathansobo:
That's a good idea. I will change it.

nathansobo:
@izuzak done.

lee-dohm:
I went ahead and fixed a broken link I found while testing the ePub version of the doc

nathansobo:
Thanks @lee-dohm.

izuzak:
:zap: @nathansobo and @lee-dohm -- looks great! :ship: (Note: I've made a few more tweaks in e72f00bdbd2dfc8354d8a877ad377f88e1a8f5fd.)

PR #62, opened 2015-05-27
Added a page regarding icons.

makerimages:
This is missing from the docs. Added it.

### DIFF
--- a/book/03-hacking-atom/sections/06-iconography.asc
+++ b/book/03-hacking-atom/sections/06-iconography.asc
@@ -1,18 +1,18 @@
 === Iconography
 
-Atom comes bundled with the [Octicons](https://octicons.github.com/) icon set. Use them to add icons to your packages.
+Atom comes bundled with the https://octicons.github.com/[Octicons] icon set. Use them to add icons to your packages.
 
 ===== Usage
 
-Atom's usage of the Octicons differs just a bit from the [standard way](https://octicons.github.com/usage/). The biggest difference is in the naming of the icon classes. Instead of the `octicon octicon-` prefix, you would use a more generic `icon icon-` prefix.
+Atom's usage of the Octicons differs just a bit from the https://octicons.github.com/usage/[standard way]. The biggest difference is in the naming of the icon classes. Instead of the `octicon octicon-` prefix, you would use a more generic `icon icon-` prefix.
 
-As an example, to add a [monitor icon](https://octicons.github.com/icon/device-desktop/), use the `icon icon-device-desktop` classes in your markup:
+As an example, to add a https://octicons.github.com/icon/device-desktop/[monitor icon], use the `icon icon-device-desktop` classes in your markup:
 
 ```html
   <span class="icon icon-device-desktop"></span>
 ```
 
-Or in case you're using [SpacePen](http://atom.github.io/space-pen/) it would be:
+Or in case you're using http://atom.github.io/space-pen/[SpacePen] it would be:
 
 ```coffeescript
   @span class: 'icon icon-device-desktop'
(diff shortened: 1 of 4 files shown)


makerimages:
Sorry, messed up the gemfile by accident. Fixed now.

makerimages:
That's a squash of the previous commits. Message not right. Sorry.

makerimages:
Anything?

/cc @izuzak

book/03-hacking-atom/sections/06-iconography.asc
```diff
@@ -0,0 +1,27 @@
...
+  icon-something
+```
+
+And so, to get the same icon to appear as a span, you could do this:
+
+```coffeescript
+  @span class: 'icon-device-desktop'
+```
+
+The result is the same: the  https://octicons.github.com/icon/device-desktop/[device-desktop] icon appears.
```
50Wliu: There's two spaces between `the` and `https:`

book/03-hacking-atom/sections/06-iconography.asc
```diff
@@ -0,0 +1,27 @@
+=== Iconography
+
+Every so often you come to a point in time, where, to represent something, an icon is better off than words are.
+Since Atom is made by GitHub, their https://octicons.github.com/[Octicons] font is available for use.
```
simurai: > icon is better off than words

Some argue that "[The best icon is a text label](http://thomasbyttebier.be/blog/the-best-icon-is-a-text-label)". So maybe we should encourage people to use text labels if possible and only if space gets tight fall back to an icon. Or often icons get used in addition to a text label.

> Since Atom is made by GitHub

Not sure about mentioning this as the main reason. I guess it is, but just sounds a bit strange in this context. How about that whole first paragraph simplified to something like:

> Atom comes bundled with the [Octicons](https://octicons.github.com/). You can use these icons in your packages if space for a text label is insufficient or to add a visual clue to an existing label.

book/03-hacking-atom/sections/06-iconography.asc
```diff
@@ -0,0 +1,27 @@
...
+
+===== Usage
+
+Atom's usage of https://octicons.github.com/[Octicons] differs just a bit from the standard usage of them, that you'd find, say, in a web project.
```
simurai: How about:

> Atom's usage of the Octicons differs just a bit from the [standard usage](https://octicons.github.com/usage/).

Since they're already linked above. Also mentioning "web project" feels not that necessary.

book/03-hacking-atom/sections/06-iconography.asc
```diff
@@ -0,0 +1,27 @@
...
+
+The biggest difference is in the naming of the icon classes. In regular webdev, you'd create something like this:
+
+```html
+  <span class = "octicon octicon-device-desktop"></span>
```
simurai: The space before and after the `=` can be removed.

simurai:
Sorry, after adding a couple of inline comments it kinda became too complex. So I made another PR https://github.com/makerimages/docs/pull/1 on top of this one that would rewrite it a bit. :innocent:

:+1: on adding this section, especially since the usage differs from the standard Octicons.

book/A-upgrading/sections/ui-theme.asc
```diff
@@ -68,8 +68,7 @@ atom-text-editor::shadow .highlight.my-linter {
 }
 ```
 
-Check out the https://github.com/atom/find-and-replace/blob/95351f261bc384960a69b66bf12eae8002da63f9/styles/find-and-replace.less#L10
-[find-and-replace]package for another example of using `::shadow` to pierce the shadow DOM.
+Check out the https://github.com/atom/find-and-replace/blob/95351f261bc384960a69b66bf12eae8002da63f9/stylesheets/find-and-replace.less#L9-L29[find-and-replace] package for another example of using `::shadow` to pierce the shadow DOM.
```
simurai: This looks like it shouldn't be part of this PR. We can still keep it since it fixes that link.

makerimages: Yeah, don't know what went wrong there :stuck_out_tongue:

makerimages:
I fixed the link markup, that @simurai broke.

simurai:
> I fixed the link markup, that @simurai broke.

:bow:

lee-dohm:
Thanks @makerimages!

PR #77, opened 2015-06-17
Add sub-section on renaming packages

mnquintana:
Adapted from https://github.com/atom/atom/blob/master/docs/apm-rest-api.md#renaming-a-package

/cc @atom/issue-triage

### DIFF
--- a/book/04-behind-atom/sections/07-maintaining-your-packages.asc
+++ b/book/04-behind-atom/sections/07-maintaining-your-packages.asc
@@ -47,6 +47,11 @@ First, edit the `name` field in your package's `package.json`:
 }
 ```
 
+Then commit your changes:
+```bash
+git commit -am "Rename package"
+```
+
 Then, publish a new version of your package:
 
 ```bash
--- a/book/04-behind-atom/sections/07-maintaining-your-packages.asc
+++ b/book/04-behind-atom/sections/07-maintaining-your-packages.asc
@@ -38,22 +38,10 @@ This will remove your package from the https://atom.io package registry. Anyone 
 
 ==== Rename Your Package
 
-You can rename your package by publishing a new version with the name changed in your package's `package.json`. Requests made to the previous name will be forwarded to the new name.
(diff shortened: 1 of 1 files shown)


book/04-behind-atom/sections/07-maintaining-your-packages.asc
```diff
@@ -35,3 +35,20 @@ apm unpublish package-name
 ```
 
 This will remove your package from the https://atom.io package registry. Anyone who has already downloaded a copy of your package will still have it and be able to use it, but it will no longer be available for installation by others.
+
+==== Rename Your Package
+
+You can rename your package by publishing a new version with the name changed in your package's `package.json`. Requests made to the previous name will forward to the new name.
```
50Wliu: `will be forwarded`?

mnquintana:
I feel like this might still be missing one step - unpublishing the package under the old name? Here's one I just renamed from [carto-atom](https://atom.io/packages/carto-atom) to [language-carto](https://atom.io/packages/language-carto). carto-atom doesn't redirect to language-carto, and still exists in the registry. Does the old package name also need to be unpublished for this to work?

/cc @thedaniel

thedaniel:
No, a rename should Just Work and redirect

mnquintana:
Hmmm, any idea why [carto-atom](https://atom.io/packages/carto-atom) isn't redirecting to [language-carto](https://atom.io/packages/language-carto)?

thedaniel:
You need to send a boolean `"rename": true` with the API call or it will create a new package. The `--rename` flag on `apm publish` does this for you, cf. `apm help publish`

mnquintana:
Derp. So is the only step actually just running `apm publish --rename new-package-name` and that's it? You don't need to push a new release first or rename anything manually?

mnquintana:
Alright, updated this with the most accurate instructions for renaming a package.

mnquintana:
How does this look now @atom/issue-triage?

PR #78, opened 2015-06-17
Convert book outline to Markdown

mnquintana:
Makes the book outline a little prettier (and lets us use things like TODO lists) :stuck_out_tongue_closed_eyes: 

/cc @atom/issue-triage


izuzak:
Looks okay to me :+1:

PR #83, opened 2015-06-25
Fixes pane title in Atom Packages section

eebs:


eebs:
Just noticed some additional out of date language, and the screen shots are out of date as well. I'll work on updating these too.

mnquintana:
Were you gonna do those in this PR? It'd probably be best to do a separate PR. Also, I'd hold off on screenshots for now - we definitely need to update them, but we're still working out standards for them. Thanks for doing this though!

eebs:
I was going to add them to this PR, but I could open a new one instead. I'm hesitant to update the wording without the screenshots, as they would no longer match. What's your preference?

eebs:
Closing, opened #84 instead.

PR #86, opened 2015-06-26
Fix link to Package Word Count from The Init File

rachelmyers:
Section [3.2: The Init File](https://atom.io/docs/latest/hacking-atom-the-init-file) includes a link to `Package: Word Count` that didn't work for me; it defaulted back to the main docs page.

(My guess is that Package Word Count content used to be on the same page, and could use the shorthand for finding a section on the same page, `<some-header>`. Since then, it seems to have moved, so I think it needs to use the longer format, `whatever_path[Text to Display]`.)

I've never written asciidoc before and was just passing through the Atom docs, so let me know if there's a better fix for this! :grin:

### DIFF
--- a/book/03-hacking-atom/sections/02-init-file.asc
+++ b/book/03-hacking-atom/sections/02-init-file.asc
@@ -1,7 +1,7 @@
 [[_the_init_file]]
 === The Init File
 
-When Atom finishes loading, it will evaluate `init.coffee` in your `~/.atom` directory, giving you a chance to run CoffeeScript code to make customizations. Code in this file has full access to https://atom.io/docs/api/latest/Atom[Atom's API]. If customizations become extensive, consider creating a package, which we will cover in <<_package_word_count>>.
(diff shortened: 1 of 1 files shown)


izuzak:
It seems that most/all such links are now busted (for example, click the "Command Palette" link in the same article).

Those links were working pre-1.0, I belive. For example, visit https://atom.io/docs/v0.207.0/hacking-atom-the-init-file and click “Package: Word Count”. Maybe something broke on the server side when Atom went 1.0? cc @benogle @thedaniel

rachelmyers:
@izuzak Ah, I didn't try enough of the links to realize most of them were broken. Is the fix to manually update all the links? :weary:

izuzak:
> Is the fix to manually update all the links? :weary:

I'm hoping :pray: that's not the fix since those same links were working in the past. In other words, this doesn't seem like a problem in the links, but with how they are processed. Maybe @schacon has some :bulb: on what's causing problems here.

izuzak:
The links are working again so I'm going to close this, @rachelmyers. Thanks for reporting this, though :bow: -- @gjtorikian and I were looking into why this was happening and it definitely still happens in the docs for version 1.0.1 of Atom: https://atom.io/docs/v1.0.1/hacking-atom-the-init-file

I'll open a new issue so that we can track this and investigate further.

PR #87, opened 2015-06-27
Correct save all opened files key shortcut

li-xinyang:
The keystroke on Mac for command window:save-all should be `alt-cmd-s` instead of `ctrl-shift-S`.

### DIFF
--- a/book/01-introduction/sections/03-basics.asc
+++ b/book/01-introduction/sections/03-basics.asc
@@ -123,7 +123,7 @@ This is a great tool if you're used to the terminal or you work from the termina
 
 Editing a file is pretty straightforward. You can click around and scroll with your mouse and type to change the content. There is no special editing mode or key commands.
 
-To save a file you can choose ``File >> Save'' from the menu bar or `cmd-S` to save the file. If you choose ``Save As'' or hit `cmd-shift-S` then you can save the current content in your editor under a different file name. Finally, you can choose `ctrl-shift-S` to save all the open files in Atom.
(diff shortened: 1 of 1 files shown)


book/01-introduction/sections/03-basics.asc
```diff
@@ -123,7 +123,7 @@ This is a great tool if you're used to the terminal or you work from the termina
 
 Editing a file is pretty straightforward. You can click around and scroll with your mouse and type to change the content. There is no special editing mode or key commands.
 
-To save a file you can choose ``File >> Save'' from the menu bar or `cmd-S` to save the file. If you choose ``Save As'' or hit `cmd-shift-S` then you can save the current content in your editor under a different file name. Finally, you can choose `ctrl-shift-S` to save all the open files in Atom.
+To save a file you can choose ``File >> Save'' from the menu bar or `cmd-S` to save the file. If you choose ``Save As'' or hit `cmd-shift-S` then you can save the current content in your editor under a different file name. Finally, you can choose `cmd-alt-S` to save all the open files in Atom.
```
izuzak: I think it'd be worth using a lowercase `s` just so users don't think they need to press `shift` as well -- see [this](https://github.com/atom/atom/blob/e008ba329fcd67c6a4f9543f8452280705602b83/keymaps/darwin.cson#L44).

izuzak:
Sorry for the delay -- this looks good, just had one comment. Happy to merge this if you can update the PR (please :bell: when you do).

book/01-introduction/sections/03-basics.asc
izuzak: > I think it'd be worth using a lowercase s just so users don't think they need to press shift as well

I fixed this up myself in 5e82e26 so that I can merge this. Thanks again

PR #90, opened 2015-07-02
Replace screenshots

simurai:
This PR replaces all screenshots with an updated 1.0 look and feel.

- [x] Getting Started
- [x] Using Atom (All except 2.1 that is part of PR https://github.com/eebs/atom-docs/pull/1)
- [x] Hacking Atom
- [x] Behind Atom
- [x] Upgrading to 1.0 APIs (Unchanged)


thomasjo:
Looks like you're using some sort of custom Finder shell (ref. book/01-introduction/images/open-file.png), perhaps it would be wise to use stock Finder to avoid confusion about missing checkbox and *New Folder* button?

mnquintana:
@thomasjo I think that is stock Finder - it's just the open dialog, rather than a regular Finder window.

thomasjo:
I got a little confused when I wrote that comment. The *New Folder* button is certainly stock, but the... wait. There checkbox shows up if one hasn't checked the "Show all file extensions" option. So this is truly stock behavior indeed! Thanks for the clarification @mnquintana, and sorry to @simurai about the confusion :bow:

simurai:
Maybe it also looked a bit odd because the window was not active.. I replaced with an active one (blue Open button). Also the colored tags are shown:

open-file

simurai:
Ok, I think this is ready.

In general it's a bit more "zoomed in", so it should be easier to see details and read text.

lee-dohm:
Thanks @simurai :grinning:

PR #91, opened 2015-07-07
Update css selector to ::shadow for .cursor

wurde:
Found this while browsing the docs for the first time. The current code example throws a deprecation warning and does nothing.  Enjoying Atom overall though, congratulations :tada:.


izuzak:
Thanks for fixing this. We have a more general issue about this (see https://github.com/atom/docs/issues/56), but just fixing this for now sounds like a good idea. :+1:

PR #92, opened 2015-07-10
:abc: Document activationHooks

joefitzgerald:
This PR documents `activationHooks`, a new key in `package.json`. See also: https://github.com/atom/atom/pull/7638

/cc @thedaniel @nathansobo


nathansobo:
How long do we want to leave this undocumented so we can experiment and make changes without disruption?

joefitzgerald:
I don't know... I mean it's just an array of strings. Would we really move to a hash?

nathansobo:
Let's live dangerously.

PR #93, opened 2015-07-14
New cover image

simurai:
Replaces the current cover image with a "retro" look.

Wasn't sure about adding the "1.0" version, but I think it's handy to quickly see what version of Atom it refers to. Especially when you download or print it.


lee-dohm:
Wow ... that's awesome! Looks good to me. Is this ready to merge?

simurai:
I haven't actually tested by "generating" the book. But the file format and dimensions are kept the same. So I think it's save. :angel:

lee-dohm:
Thanks @simurai :grinning:

PR #97, opened 2015-07-21
Update Atom Flight Manual chapter 2

UsabilityEtc:
Fix typos, correct spelling and improve grammar and sentence structure in chapter 2 of the Atom Flight Manual.

### DIFF
--- a/book/02-using-atom/sections/06-customizing.asc
+++ b/book/02-using-atom/sections/06-customizing.asc
@@ -85,7 +85,7 @@ You can open this file in an editor from the _Atom > Open Your Config_ menu.
 ** `nonWordCharacters`: A string of non-word characters to define word boundaries
 ** `fontSize`: The editor font size
 ** `fontFamily`: The editor font family
-** `invisibles`: Specify characters that Atom renders as invisible in this hash (use `false` to turn off individual character types)
+** `invisibles`: A hash of characters Atom will use to render whitespace characters. Keys are whitespace character types, values are rendered characters (use value `false` to turn off individual whitespace character types)
 *** `tab`: Hard tab characters
 *** `cr`: Carriage return (for Microsoft-style line endings)
 *** `eol`: `\n` characters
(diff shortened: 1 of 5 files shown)


book/02-using-atom/sections/06-customizing.asc
```diff
@@ -85,7 +85,7 @@ You can open this file in an editor from the _Atom > Open Your Config_ menu.
 ** `nonWordCharacters`: A string of non-word characters to define word boundaries
 ** `fontSize`: The editor font size
 ** `fontFamily`: The editor font family
-** `invisibles`: Specify characters that Atom renders for invisibles in this hash (use `false` to turn off individual character types)
+** `invisibles`: Specify characters that Atom renders as invisible in this hash (use `false` to turn off individual character types)
```
izuzak: The purpose of the `invisibles` object is to specify how whitespace characters (such as spaces, tabs, and newlines) will be rendered. For example:

```
  editor:
    invisibles:
      space: "O"
```

makes space characters show up as "O":

screen shot 2015-07-22 at 16 56 16

Does that match the meaning of the sentence after your changes? To me, it sounds like you're saying that these characters will be made invisible by the editor, which isn't the same thing (and is not true).

izuzak:
This is amazing, @UsabilityEtc -- thank you! :sparkles: :memo: :sparkles:  I had just one comment, but otherwise looking forward to merging this.

book/02-using-atom/sections/06-customizing.asc
UsabilityEtc: You're right, @izuzak. I actually need to rewrite the definition of `invisibles` to better explain the key/value relationship in the hash, which is that the whitespace character (the key) will be replaced with the value.

izuzak: > I actually need to rewrite the definition of invisibles to better explain the key/value relationship in the hash, which is that the whitespace character (the key) will be replaced with the value.

:heart_eyes:

UsabilityEtc:
Thanks for the feedback, @izuzak.

izuzak:
:zap: Thanks again, @UsabilityEtc!

PR #99, opened 2015-07-23
Fix broken link and spelling error in Flight Manual chapter 1

UsabilityEtc:
Move the location of the link target `[[_installing_atom_on_mac]]` in `02-installing.asc`.

The current location of `[[_installing_atom_on_mac]]` works when the book is built locally and viewed as a single web page (`atom.html`) but is broken when the book's HTML is split up into files on https://atom.io/docs/v1.0.2/. The new location in this commit seems to produce the correct HTML that should fix the problem.

Also fix a spelling error.


izuzak:
> The new location in this commit seems to produce the correct HTML that should fix the problem.

:pray: :smile: 

Thanks again for the great fixes!

PR #105, opened 2015-08-04
autocomplete was replaced with autocomplete-plus

stevemao:


thomasjo:
Thanks for helping out! Would you be open to rewriting the rest of the section as well? It's a little bit outdated, so we really need to fix the entire thing, including the screenshot.

stevemao:
Sure. But the screenshot looks fine to me? There is not many places I should rewrite? Is it looking good?

thomasjo:
> But the screenshot looks fine to me?

Oops, you're right. I must have looked at an older version.

> There is not many places I should rewrite?

The paragraph related to the setting for searching all buffers in incorrect, since this is the default in autocomplete-plus. We should probably remove or rephrase that paragraph. And the paragraph preceding that one should also be rephrased to state that by default all currently open files are searched.

/cc @izuzak @atom/feedback

book/02-using-atom/sections/02-text.asc
```diff
@@ -338,7 +338,7 @@ image::images/autocomplete.png[autocomplete]
 
 By default, the autocompleter will look through the current open file for strings that match what you're starting to type.
 
-If you want more options, in the Settings panel for the Autocomplete package you can toggle a setting to make the autocompleter look for strings in all your open buffers rather than just the current file.
+If you want more options, in the Settings panel for the Autocomplete-plus package you can toggle a setting to make the autocompleter look for strings in all your open buffers rather than just the current file.
```
thomasjo: We should either refer to the package as Autocomplete Plus or autocomplete-plus, since those are the two naming conventions used in Atom (including the settings-view package).

mnquintana: I think we should go with autocomplete-plus - the humanized name is only used for the breadcrumb in settings-view now and I was thinking about removing it.

stevemao:
Hi guys, have we reached a consensus on this matter?

book/02-using-atom/sections/02-text.asc
izuzak: :+1: on using `autocomplete-plus`, at least for now.

izuzak:
> Hi guys, have we reached a consensus on this matter?

@stevemao Just to make sure I understand -- can you clarify which questions you're blocked on? I just :+1: one of the inline comments, not sure if there's anything else here you're waiting on.

stevemao:
I mean what else do I need to do before this is merged? Seems like @thomasjo and @thomasjo have argument on "Autocomplete Plus" or "autocomplete-plus"?

thomasjo:
There's most certainly no argument :smile: I was fine with either alternative, and thus we now have 3 votes in favor of *autocomplete-plus*, so change it to that, and we're good to go!

stevemao:
Done. Thanks :)

thomasjo:
We still need to rewrite this section since it's not really correct given the differences between the old `autocomplete` and the current `autocomplete-plus`, but I'm merging this regardless. We can improve the rest in a separate PR.

thomasjo:
Thanks for the help @stevemao :bow:

PR #114, opened 2015-08-26
[WIP]  Docs on scopes

MaximSokolov:
atom/atom#8430
No new scopes yet. Just standart TextMate root groups.
No descripttion yet.
I think definitions like `storage — things relating to “storage”` isn't goog enough, so I added a couple examples


simurai:
:+1: Having an official grammar naming convention would be great. Then there is at least a place to refer to. And the convention could be refined over time.

How much does this version differ from the TextMate convention? I think until an Atom 2.0, it would need to stay somewhat close. Just because there are already lots of themes out there that are mostly based on TextMate's convention. Smaller changes are fine, just not big ones.

MaximSokolov:
This PR just a skeleton of convention. **Don't merge it yet**.

TextMate docs are unclear. Seems like people just look at list of scopes and don't read text above. So I think in docs should be less text and more examples. approach redundant

> How much does this version differ from the TextMate convention?

You can view all new scopes in the issue above. I added a couple optional scopes. (e.g. `keyword.operator` -> `keyword.operator.comparison`) The main difference is a `punctuation` scope. Most of suggested scopes already is used in languages. However it still would be very hard to make general languages follow convention and don't break syntax highlighting.

MaximSokolov:
> - meta
  - function
  - function-call
  - method-call

Scopes currently is used:
* `meta.function` in coffee, js, objective-c, perl, php, python, ruby, c/c++...
* `function-call` in php, python
* `method-call` in c#, java

Suppose I have this pseudocode:
```js
function functionName(param1, param2) {
  ...
}
```
And I want to change color for `functionName`:
```less
.entity.name.function {
  color: @someColor;
}
```
That's how all syntax themes highlight `functionName`. Let's look at TextMate docs:

> entity — an entity refers to a larger part of the document, for example ... a function ...
  ...
  * name — we are naming the larger entity.
      * function — the name of a function.

> ...

However according this definition `entity.name.function` should also highlight `functionCall` and `methodCall`:
```js
functionCall(arg1, arg2)
obj.methodCall(arg1, arg2)
```
How to change color for `functionCall` and `methodCall`? We can use another scope for `functionCall` and `methodCall` like in Java (`meta.method`) This approach wrong because:
1.  `functionCall` and `methodCall` are definitely `entity.name.function` according definition
2. `meta` scope shouldn't be highlighted

That's why we need `meta.function`, `meta.function-call`, `meta.method-call`:
```less
.meta.function {
  .entity.name.function {
      color: @color1;
  }
}

.meta.function-call {
  .entity.name.function {
      color: @color2;
  }
}

.meta.method-call {
  .entity.name.function {
      color: @color3;
  }
}
```

For example I don't want to highlight function/method calls:
```less
.function.name {
    color: @someColor;
}

.function-call, .method-call {
  .function.name. {
      color: @syntax-text-color;
  }
}
```
I think it's non-obvious and we should put it the documentation. Of course if you agree with examples above. I will write about `parameters`, `arguments` scopes later.

MaximSokolov:
I think these scopes are redundant:

>
comment 
* line
  * **double-slash**
  * **double-dash**
  * **number-sign**
  * **percentage**
  * **character**

> 
string
* **quoted**
  * **single**
  * **double**
  * **triple**
  * **other**
* **unquoted**

These scopes don't append any information about language symbol.
I doubt someone would use different color for `'string'` and `"string"`

MaximSokolov:
> 
markup
* **changed**
* **inserted**
* **deleted**

Already is used in language-git, diff grammar from [akonwi/git-plus](https://github.com/akonwi/git-plus) package.

> 
markup
* **code**
* **strike**

Already is used in gfm

>
markup
* raw
  * **inline**
  * **block**

From literate coffeescript

mnquintana:
> I doubt someone would use different color for 'string' and "string"

That may be true, but scopes are used for far more than just syntax highlighting - since they're what give raw text semantic structure in Atom, they're also used for things like autocomplete and static analysis. So while the distinction between single- and double-quoted strings might not really matter for syntax highlighting, it could make all the difference in other contexts.

MaximSokolov:
@mnquintana Agree. However we can just get punctuation via `string` -> `punctuation.definition`

What do you think about:
>
##### Remove `begin`, `end`

>
TextMate has `begin`, `end` subscopes. Probably for brackets highlighting.
Atom doesn't need it (does it :question:):
`punctuation.definition.string.begin` -> `punctuation.definition.string`
`punctuation.definition.string.end` -> `punctuation.definition.string`

Should we remove these scopes?

MaximSokolov:
Seems like [Bracket Matcher](https://github.com/atom/bracket-matcher) package don't use these scopes. Other packages can use Bracket Matcher API's.

chbk:
Where would brackets, braces and parentheses go? How about a "section" category in "punctuation", I like the way language-ruby handles those characters.

MaximSokolov:
> Where would brackets, braces and parentheses go? How about a "section" category in "punctuation", I like the way language-ruby handles those characters.

IMO `punctuation.definition` scope makes more sense for this purpose:
```js
// `punctuation.definition.parameters` for `(`, `)`
test(param, param) {
  // `punctuation.definition.function[.brace.curly]` for `{`, `}`
  ...
}
```
`punctuation.definition.array` for `[`, `]`
`punctuation.definition.parameters`
`punctuation.definition.methods`
etc.
Even `punctuation.definition.section`

chbk:
That makes sense, I wasn't sure how you interpreted `punctuation.definition` since you only talked about strings and comments in your original post.

MaximSokolov:
> That makes sense, I wasn't sure how you interpreted punctuation.definition since you only talked about strings and comments in your original post.

WIP :) I din't even add examples for standart TextMate scopes yet.

lee-dohm:
Is this up-to-date with all the changes that the @atom/languages team have been doing?

alexchandel:
@MaximSokolov In addition to documenting these scopes, could you document how these scopes can be debugged for a particular grammar (e.g. when looking at a file highlighted with that grammar, when trying to find out why a particular scope was applied to a string of text, etc...)? The [Hacking Atom: Debugging](https://atom.io/docs/latest/hacking-atom-debugging) has laughably little to say on this.

infininight:
@MaximSokolov Didn't see this until now, we never added the punctuation scope to the documentation but we do have a spec I've [posted here for reference](https://gist.github.com/infininight/9392d1e4408d507a4691e3daba9b3ec9). This is what we go by for the TextMate grammars.

This isn't the most complete list but there is also completion for scopes when editing grammars in TextMate, the list of scopes completed is in [this command](https://github.com/textmate/bundle-development.tmbundle/blob/master/Commands/Scope%20Completion.tmCommand).

\- Michael Sheets, TextMate Bundle Maintainer

lolorz13:
I noticed that this does not reference the entity scope in the textmate doc, is this intentional? for python this would be useful for "import somepackage" where "somepackage" would be entity.name.package

tabathap:
👌

PR #119, opened 2015-08-28
Squash outline.txt to outline.md

Lorwp:


mnquintana:
Thanks for fixing this! :boat:

PR #34, opened 2015-03-21
Update LICENSE.asc: change license to CC BY-SA 3.0

Calinou:
As seen here: https://github.com/atom/docs/issues/33

PR #122, opened 2015-09-02
Fix Wrong License in Readme.md

Buddybenj:
See PR #34


mnquintana:
Nice catch! Thanks for fixing this

PR #129, opened 2015-10-05
Add Timecop section to debugging guide and tweak some other sections

izuzak:
This replaces a `TODO` in the debugging guide to write something about Timecop with some actual text. I also updated some other sections here and there (some text was outdated), and re-arranged the sections so that the more common/useful stuff is nearer to the top.

@atom/feedback If you have any thoughts or suggestions -- please let me know. :pray: The list of commits might be useful here for review because the total diff is a bit large due to the re-arranging of sections. I plan on merging this near the end of the week. :calendar:


thedaniel:
I left one small comment but I think this is great.

PR #132, opened 2015-10-12
Fix typo in link in section 3.2

### LINKED ISSUE #123: Typo in a link near the bottom of Section 3.2: The Init File
In the first sentence of the last paragraph:

> Now, reload Atom and use the **[Section 1-9](https://atom.io/docs/latest/1-introduction/_command_palette)** to execute 
> the new command by name (i.e., "Markdown: Paste As Link").

I believe it should should say "...and use the *command palette* to execute..." instead. Also, the link is broken, and the URL it points to returns a 404 error.

formap:
I fixed the link so it would point to the command palette repo. I wasn't sure if it should link to where i linked it or [here] (https://atom.io/docs/latest/getting-started-atom-basics#command-palette) but i was hoping you guys could give me feedback and i can amend the commit if necessary. Thanks!

Closes #123.

### DIFF
--- a/book/03-hacking-atom/sections/02-init-file.asc
+++ b/book/03-hacking-atom/sections/02-init-file.asc
@@ -23,4 +23,4 @@ atom.commands.add 'atom-text-editor', 'markdown:paste-as-link', ->
   selection.insertText("[#{selection.getText()}](#{clipboardText})")
 ```
 
-Now, reload Atom and use the <<_command_palette>> to execute the new command by name (i.e., "Markdown: Paste As Link"). And if you'd like to trigger the command via a keyboard shortcut, you can define a <<_customizing_keybindings,keymap for the command>>.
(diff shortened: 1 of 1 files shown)


izuzak:
Thanks for opening a pull request, @formap. 

Yeah, that text should link to the command palette section here:

https://atom.io/docs/v1.0.19/getting-started-atom-basics#command-palette

Previously, that link worked, and I'm not sure why it doesn't work anymore. There seem to be two problems: 1) the link doesn't work actually, 2) the text for the link isn't "Command Palette" (which it used to be before). I'm guessing this is another problem with the platform we use to generate the docs (Atlas). 

It seems that all links of the form `<<_section_name>>` are broken now. I think we should figure out why and fix that problem instead of changing all links individually to use absolute URLs. 

/cc @thedaniel

PR #136, opened 2015-10-20
Add note on --profile-startup flag to debugging guide

izuzak:
cc @kevinsawicki for :eyes: because you added this in https://github.com/atom/atom/pull/6543 and I think it would be neat to have it documented here so that we can link to it.

### DIFF
--- a/book/03-hacking-atom/sections/A01-debugging.asc
+++ b/book/03-hacking-atom/sections/A01-debugging.asc
@@ -116,7 +116,7 @@ image::images/devtools-error.png[devtools error]
 
 If you can reproduce the error, use this approach to get the full stack trace and https://github.com/atom/atom/blob/master/CONTRIBUTING.md#submitting-issues[report the issue].
 
-==== Diagnose startup performance problems with the Timecop package
+==== Diagnose startup performance problems with the Timecop package and the `--profile-startup` command line flag
 
 If Atom is taking a long time to start, you can use the https://github.com/atom/timecop[Timecop package] to get insight into where Atom spends time while loading.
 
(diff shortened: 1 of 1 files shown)


kevinsawicki:
Looks good, thanks for adding this :+1:

izuzak:
:bow:

@atom/issue-triage @atom/feedback Friendly ping on this, in case you need to link to such instructions.

nathansobo:
One thing worth noting is that it only seems to work correctly when Atom has already started. Should we investigate that before merging this so as not to generate confusion?

izuzak:
> One thing worth noting is that it only seems to work correctly when Atom has already started. Should we investigate that before merging this so as not to generate confusion?

Hmmm, is that true? It works fine for me even if I don't have an Atom window open.

izuzak:
Ohhhh wait -- it seems to break on the second run. So, after I first quit Atom, I could trigger the profile with the `--profile-startup` flag. But after I quit again and started Atom again with that flag -- it didn't do its thing. It did open the devtools, but no profile created.

izuzak:
@nathansobo @kevinsawicki I'm not super familiar with the profiles tab in devtools -- is that :point_up:  by any chance expected behavior or would that be a problem in Atom/Electron?

kevinsawicki:
Hmm, I wonder if the timeout needs to be bumped in https://github.com/atom/atom/blob/806b46bb6c981f110a34ed20f566de50288900d0/static/index.js#L143-L145

There seemed to be issues in past Electron versions where you needed to wait a bit to start profiling a  after the devtools opened event fired.

izuzak:
> Hmm, I wonder if the timeout needs to be bumped in https://github.com/atom/atom/blob/806b46bb6c981f110a34ed20f566de50288900d0/static/index.js#L143-L145

@kevinsawicki I tried bumping the timeout to 500ms and it did resolve the problem for me. :tada: 

I'm not sure what a good timeout would be, so I'll just open an issue to track this and perhaps you or @nathansobo can pick a good timeout value? :hourglass:

nathansobo:
@izuzak It always works reliably for me once Atom is started, but if I completely quit out of Atom and launch it for the first time using this flag, it doesn't work. Trying to investigate...

PR #138, opened 2015-10-23
Update iconography section

simurai:
:warning: This PR is dependant on https://github.com/atom/styleguide/pull/32

Now that Atom's version of the Octicons isn't up to date with https://octicons.github.com, it's probably a good idea to not link to it. The new ones look a bit different and some icons got removed or new ones added. 

This PR updates the copy to point to the Styleguide to get an overview of all icons.


lee-dohm:
Looks good to me. Are we going to update to the latest version of Octicons at some point?

simurai:
> Are we going to update to the latest version of Octicons at some point?

We could, but some icons got removed: https://github.com/atom/atom/issues/8314 and would break the "API". Maybe some kind of custom generated version with old and new ones mixed.

PR #149, opened 2015-12-15
Update atom versions and clarify repository URL

gimenete:


lee-dohm:
Thanks for contributing!

PR #154, opened 2016-01-06
Update 06-iconography.asc

gsklee:
Remove information about SpacePen since it's been deprecated.


simurai:
There is also #138 which might cause a conflict when trying to merge both.

simurai:
Resolved the conflict in #169 and reference to SpacePen will be removed.

PR #156, opened 2016-01-07
Update 03-scopes.asc

gsklee:
Update `atom.config.set()` usage example to feature the latest API.


mnquintana:
Thanks for fixing this! :rocket:

PR #157, opened 2016-01-11
Fix #85, Add a warning note to Word Count tutorial

### LINKED ISSUE #85: Package name in first package writing tutorial causes issue with not being loaded
See: atom/settings-view#554

5t111111:
To avoid confusion mentioned in #85, I've Added a waring note to "Package: Word Count" tutorial.


lee-dohm:
Thanks for your help!

PR #168, opened 2016-03-03
Add sentence on Keybindings Resolver view.

eldondev:
Because it can be hard to tell what other, more-specific selector is
overriding a user customization. In my case, it was a platform-specific
selector.


lee-dohm:
Thanks for this :grinning:

PR #170, opened 2016-03-09
Document portable mode for Windows

damieng:


50Wliu:
Closes atom/atom#10072.

PR #171, opened 2016-03-14
Add summary chapter 3

roebenk:
Added a short summary to chapter 3, because the [current page](https://atom.io/docs/v1.5.3/hacking-atom-summary) was still empty. This will look a lot better and give more info.

### DIFF
--- a/book/03-hacking-atom/1-hacking-atom.asc
+++ b/book/03-hacking-atom/1-hacking-atom.asc
@@ -27,6 +27,8 @@ include::sections/A03-textmate.asc[]
 
 === Summary
 
-( List of topics we covered / appendix? )
+If you finished this chapter, you should be an Atom-hacking master. We've discussed how you should work with CoffeeScript, and how to put it to good use in creating packages. You should also be able to do this in your own created theme now.
 
-Summary.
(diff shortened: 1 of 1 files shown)


book/03-hacking-atom/1-hacking-atom.asc
```diff
@@ -27,6 +27,8 @@ include::sections/A03-textmate.asc[]
 
 === Summary
 
-( List of topics we covered / appendix? )
+If you finished this chapter, you should be an Atom-hacking master. We've discussed how you should work with CoffeeScript, and how to put it to good use in creating packages. You should also be able to do this in your own created theme now.
```
lee-dohm: In summaries such as this, I tend to avoid "you should" statements. If the manual didn't explain things well enough, it may be the fault of the manual, not the reader. "You should" statements imply that the fault lies with the reader.

I'd recommend sticking to "we discussed", "we outlined", "we described" statements. Example: "We discussed how to work with CoffeeScript and how it can be used in creating packages and themes."

lee-dohm: Actually, I've been reviewing the docs quite a bit over the past couple days and this appears to be the same style as the others :grinning:

lee-dohm:
Thanks for this!

PR #174, opened 2016-03-20
Fix grammar in 05-themes.asc

houseoftech:


lee-dohm:
Thanks for this! I updated it for the new static site architecture.

PR #179, opened 2016-03-24
Clearer list of activationHooks

rahatarmanahmed:
The list of known activation hooks was previously confusing because the list had only one item, and when rendered showed up as a bullet point in the same level list of `package.json` properties.

The only hook I could find any mention of was the one that was already there, so I assume it's the only one. Changed the wording to indicate that.

Also made the use of e.g. clearer w/ parentheses.


thedaniel:
This is great clarification, and searching the code for use of `triggerActivationHook` does indicate that `grammar-used` is the only one, but I think the language "only currently known hook" makes it seem like a mystery. I think it's better just to say something like "currently, the only activation hook is"

rahatarmanahmed:
Updated the PR with the wording change.

thedaniel:
CI failure looks unrelated, so I am merging. Thank you very much for contributing!

PR #184, opened 2016-03-27
Switch to modern smart quote syntax

mojavelinux:
See http://asciidoctor.org/docs/migration/#migration-cheatsheet for details. Asciidoctor is now using the modern syntax everywhere by default (unless you explicitly enable compat-mode).


gjtorikian:
Thanks!

PR #188, opened 2016-04-11
Fix typo in service doc

deprint:
Found a small syntax error in [Chapter 4: Interacting With Other Packages Via Services](http://flight-manual.atom.io/behind-atom/sections/interacting-with-other-packages-via-services/):

screenshot_2016-04-11_14-04-39


50Wliu:
:+1:

PR #191, opened 2016-04-24
Fix figure title

mzlogin:
This figure's title and content are not match.


lee-dohm:
Thanks very much for contributing!

PR #221, opened 2016-07-11
Windows build compatibility via Ruby 2.3.0

damieng:


.ruby-version
```diff
@@ -1 +1 @@
-2.2.3
+2.2.4
```
lee-dohm: Did you mean for this to be `2.3.0` also?

damieng: Oops.

lee-dohm:
Works fine! Thanks :bow:

lee-dohm:
It turns out that the auto-publish infrastructure we're using doesn't support Ruby v2.3.0 right now #222 ... reverting to unblock auto-publishing until we can figure out the right solution.

PR #226, opened 2016-07-13
Don't upgrade to nanoc v4.2

### LINKED ISSUE #205: nanoc crashes with V4.2.0
nanoc >=v4.2.0 is not compatible with nanoc-conref-fs any more.
if nanoc >=4.2 is installed, the build is broken

suggesting change in gemfile to fix the issue.

lee-dohm:
Fixes #205

Since the version of nanoc we're currently using works fine for our purposes, we'll stick with v4.1.x for now until gjtorikian/nanoc-conref-fs#5 is fixed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,8 @@ source 'https://rubygems.org'
 ruby '2.2.3'
 
 # essential
-gem 'nanoc', '~> 4.1'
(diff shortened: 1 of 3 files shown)


gjtorikian:
https://github.com/gjtorikian/nanoc-conref-fs/pull/8 released 0.6.6 which fixed the issue.

PR #177, opened 2016-03-22
Update 02-installing.asc

fusion809:
This pull request corresponds to issue #175. It does **NOT** include mention of the unofficial, distribution-specific, means of installing Atom on Linux. Like from the Arch User Repository, `sabayon` Gentoo overlay, *etc.*

PR #227, opened 2016-07-13
Add extra detail to Linux installation instructions

lee-dohm:
Update #177 from @fusion809 to the new format and include @thedaniel's feedback.

### DIFF
--- a/content/getting-started/sections/installing-atom.md
+++ b/content/getting-started/sections/installing-atom.md
@@ -93,20 +93,50 @@ Should the user uninstall Atom it will only affect themselves and it will not au
 
 #### Installing Atom on Linux
 
-To install Atom on Linux, you can download a [Debian package](https://atom.io/download/deb) or [RPM package](https://atom.io/download/rpm) either from the [main Atom website](https://atom.io) or from the [Atom project releases page][releases].
+To install Atom on Linux, you can download a [Debian package](https://atom.io/download/deb) or [RPM package](https://atom.io/download/rpm) either from the [main Atom website](https://atom.io) or from the [Atom project releases page][releases]. These packages do not currently have auto-update features, so when you would like to upgrade to a new release of Atom, you will have to repeat this installation process.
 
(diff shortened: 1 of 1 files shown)


fusion809:
Ah https://github.com/atom/atom/archive/atom.x86_64.rpm does not exist (sorry if I am the one that make this mistake, as I checked #177 and I seemed to have made this stuff up too)... To download the latest RPM release you would use the URL https://atom.io/download/rpm (e.g., `curl -L https://atom.io/download/rpm > atom.x86_64.rpm`). This won't work if you pass it to the package managers though, as they don't recognize it without a .rpm file extension (although the full link like https://github.com/atom/atom/archive/v1.8.0/atom.x86_64.rpm would work when passed to `dnf`, `yum` and `zypper`).

lee-dohm:
Thanks for the info, I'll see about fixing it.

fusion809:
By the way, why doesn't the documentation mention the Linux tarball releases? Some users would probably refer to use a binary tarball to run Atom as they don't need to install it.

lee-dohm:
@fusion809 Please feel free to submit a PR with some instructions for it. I'm not familiar enough with the process. I'll be working on the manual as my primary effort for the next week or so. If you can get the PR submitted in that time frame I can review and merge it right away.

fusion809:
Sure, but to do this I would need a release-independent link to download the tarball from. I have tried https://atom.io/download/tar, https://atom.io/download/targz, https://atom.io/download/tar.gz, https://atom.io/download/tarball, https://atom.io/download/linux, *etc.* But none of these links to the latest `atom-amd64.tar.gz` binary tarball.

lee-dohm:
I would guess that's why we don't have instructions on that :grinning:

fusion809:
So do you plan to fix that or get someone else to fix it? I ask because if not it may not be worthwhile me writing these instructions.

lee-dohm:
We're not currently planning on adding it, no. So yeah, it would be a waste to add the instructions right now.

PR #231, opened 2016-07-13
Fix oversized footer on root page

lee-dohm:
Added an extra guard to ensure that the TOC style doesn't bleed over into the footer.

/cc @simurai Not sure if this is the best way to do this, though it appears effective. 💭 ?


simurai:
Yeah, I think this good. Let's 🐑 it.

PR #240, opened 2016-07-19
Fix typo

brandoncc:


mnquintana:
Nice catch, thanks for fixing this! 🚀

PR #250, opened 2016-08-04
Restyle platform-nav

### LINKED ISSUE #249: Restyle platform picker
How about moving the "platform picker" next to the search input? Maybe like:

screen shot 2016-08-03 at 10 14 55 am

- It would make it look more like a global setting and not something you would change for every page.
- It would give a bit more breathing room to the title + "Improve this page"

My only concern is that it might look like it's part of the search. Like if you type something and then click one of the platform buttons and nothing happens. A separator line might make it more clear:

screen shot 2016-08-03 at 10 28 59 am

/cc @lee-dohm

simurai:
This
- removes the article title (already shown)
- moves the platform-nav next to the "Improve this page" link
- restyles the platform-nav

Before | After
--- | ---
screen shot 2016-08-04 at 5 37 01 pm | screen shot 2016-08-04 at 5 36 46 pm

screen shot 2016-08-04 at 5 36 14 pm

Closes #249

### DIFF
--- a/assets/javascripts/platform-nav.js
+++ b/assets/javascripts/platform-nav.js
@@ -48,7 +48,7 @@
 
   window.initPlatformNav = function () {
     var $platformNav = $('#platform-nav')
-    $platformNav.show()
+    $platformNav.addClass('show')
     $platformNav.find('a').click(function () {
       updateArticlePlatform($(this).data('platform'))
     })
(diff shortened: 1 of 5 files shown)


bastilian:
Looks awesome!

lee-dohm:
Thanks for making things look awesome :grinning:

PR #252, opened 2016-08-05
Update editing-and-deleting-text.md

zimenskiy:
fix a typo


lee-dohm:
Thanks for catching this! Multiple people submitted the same correction and I missed that you caught it first. Thank you very much for helping out!

PR #254, opened 2016-08-09
Note on "**" support in glob patterns

colinmorris:
(Surprisingly, globstar is not actually mentioned in the linked Wikipedia article)


lee-dohm:
Thanks for contributing!

PR #255, opened 2016-08-12
In linux move to the top/bottom of the file - is Ctrl+home/end

dspdesign:
Do not know how it on mac or windows


50Wliu:
It's the same on Windows.  Thanks for the contribution!

PR #259, opened 2016-08-25
Added example of multiple snippets per scope.

jmssmth:
Wanted to make it a bit more obvious how to create multiple snippets per scope for others who are as unfamiliar with CSON as I am. The vague warning (and link) to *Configuring with CSON* didn't help either, so after some trial and error, I figured this out. Shortly thereafter found the *Configuring with CSON* section that confirmed what I had come across.


lee-dohm:
Thanks for contributing!

PR #261, opened 2016-08-29
Added note about Git repository root

mivanov:
Until https://github.com/atom/atom/issues/2203 is resolved Git integration only works if the project root contains the repository. This took me a couple of hours to figure out because all the docs seem to say Git integration is built-in and should just work.


lee-dohm:
Thanks for contributing!

Issue #253, opened 2016-08-08
Search not working


lee-dohm:
The search functionality doesn't seem to be working anymore :cry:

BinaryMuse:
Fixed by #263, thanks @bastilian

PR #263, opened 2016-08-30
Include markdown files into search.json

bastilian:
Markdown files are currently left out when building `search.json` resulting in #253.
This adds them again and brings back search functionality.


BinaryMuse:
Thanks so much!

PR #267, opened 2016-09-13
Update atom-basics.md

ndavidow:
It's important to mention that vim and emacs styles are supported by packages


lee-dohm:
Thanks for the contribution. The change has been updated and submitted as part of #287.

PR #274, opened 2016-10-02
Minor grammatical fixes to developing-node-modules

50Wliu:
Self-explanatory.

### DIFF
--- a/content/behind-atom/sections/developing-node-modules.md
+++ b/content/behind-atom/sections/developing-node-modules.md
@@ -7,7 +7,7 @@ Atom contains a number of packages that are Node modules instead of Atom package
 
 #### Linking a Node Module Into Your Atom Dev Environment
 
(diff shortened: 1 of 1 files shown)


content/behind-atom/sections/developing-node-modules.md
```diff
@@ -7,15 +7,15 @@ Atom contains a number of packages that are Node modules instead of Atom package
 
 #### Linking a Node Module Into Your Atom Dev Environment
 
-Here are the steps to run a local version of a node module *not an apm* within Atom. We're using `atom-keymap` as an example:
+Here are the steps to run a local version of a node module within Atom. We're using `atom-keymap` as an example:
```
BinaryMuse: I think the "not an apm" wording was weird and awkward, but I wonder if it's worth highlighting that we're talking about a Node module here, and not an Atom package?

lee-dohm:
Thanks @50Wliu, I've superseded this with #279 and addressed @BinaryMuse's comment too.

PR #282, opened 2016-10-23
Update editing-and-deleting-text.md

realtime:
The "Deleting and Cutting" section only describes delete operations, so "Cutting" was removed.


lee-dohm:
Thanks for contributing! There is an [OS-specific section](https://github.com/atom/flight-manual.atom.io/blob/master/CONTRIBUTING.md#os-specific-sections) where a keybinding for cutting text is mentioned. Even though it may be slightly confusing, I feel the title is best left the way it is.

Thanks again for taking the time to help out :grinning:

PR #284, opened 2016-10-28
Update language-scoped settings list

DSpeckhals:
There are currently a few missing settings from this list, and one removed. The current list can be found at https://github.com/atom/settings-view/blob/master/lib/settings-panel.coffee#L17. The removed setting has a PR open here: https://github.com/atom/settings-view/issues/873.

I also put the items in order of appearance in the settings view (alphabetic).

### DIFF
--- a/content/using-atom/sections/basic-customization.md
+++ b/content/using-atom/sections/basic-customization.md
@@ -173,17 +173,18 @@ You can also set several configuration settings differently for different file t
 There are several settings now scoped to an editor's language. Here is the current list:
 
 ```
-editor.tabLength
-editor.softWrap
-editor.softWrapAtPreferredLineLength
+editor.autoIndent
+editor.autoIndentOnPaste
+editor.invisibles
+editor.nonWordCharacters
 editor.preferredLineLength
 editor.scrollPastEnd
-editor.showInvisibles
 editor.showIndentGuide
-editor.nonWordCharacters
-editor.invisibles
-editor.autoIndent
-editor.normalizeIndentOnPaste
+editor.showInvisibles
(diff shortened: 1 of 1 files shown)


50Wliu:
Can you please fix the merge conflicts @DSpeckhals?

DSpeckhals:
This is fixed.

50Wliu:
I'm pretty sure this isn't the entire list because `editor.softTabs` and `editor.tabType` can also be scoped.

DSpeckhals:
I see what you mean. It appears, according to where this is at in the document, that it applies to _all_ scoped settings; not just ones exposed in the Settings View. Do you think we should display all of the settings, or just the ones in the Settings View?

The one problem that I have is that if we label all 20+ settings here, it might be a bit overwhelming.

Here's the list that I see in Atom Core's config schema:
```
fontFamily
fontSize
lineHeight
showInvisibles
showIndentGuide
showLineNumbers
atomicSoftTabs
autoIndent
autoIndentOnPaste
nonWordCharacters
preferredLineLength
tabLength
softWrap
softTabs
tabType
softWrapAtPreferredLineLength
softWrapHangingIndent
scrollSensitivity
scrollPastEnd
undoGroupingInterval
confirmCheckoutHeadRevision
invisibles
zoomFontWhenCtrlScrolling
```

lee-dohm:
Thanks for contributing!

Long term, we should come up with a way for users to discover the full set of language-scoped settings either through the application or someplace we can link to directly in the code. Ultimately, we shouldn't have to maintain this list separately in both the code and in the Flight Manual.

PR #297, opened 2016-11-21
Update package-word-count.md

maveme:


50Wliu:
The default keymap is in CSON format.

PR #299, opened 2016-11-29
Clarify what to return in package deserialization

BinaryMuse:
It says to return JSON, but it's unclear if one should return a serialized object (a string) or an object to be serialized.


kuychaco:
👍

PR #300, opened 2016-11-30
Document `initialize()` hook

matthewwithanm:
Just a little documentation for the new `initialize()` hook: what it is and when to use it.

Released under CC0.


nathansobo:
@lee-dohm This is document a feature that isn't even merged to master yet. Is there some facility for versioning the flight manual or should we have held off on merging this?

lee-dohm:
@nathansobo There isn't a facility for versioning the flight manual, no. I probably should have held off on merging it. I'll try to do better in the future :grinning:

PR #301, opened 2016-11-30
Document registering deserializers in package.json

matthewwithanm:
Some documentation for the (relatively) new ability to register deserializers via `package.json`.

Released under CC0.


nathansobo:
Thanks for going the extra mile on this! ⚡️

PR #311, opened 2017-01-03
Change keybinding for Developer Tools on Windows

barik:
The keybinding appears to be Ctrl+Shift+I for Windows also, not Alt+Ctrl+I.


50Wliu:
Thanks!  I'm going to make an extra change on master to combine the two classes.

PR #312, opened 2017-01-05
Remove alt-shift-s reference

theredcoder:
The default keybinding was removed in atom/snippets.  Ref: atom/snippets@690d4fd


theredcoder:
@50Wliu Thanks for the quick feedback.  The suggested changes are now included in the PR.

lee-dohm:
Thanks for the contribution!

PR #315, opened 2017-01-19
French Version of the chapter 1: Why Atom?

Bruneton:
I would like to add a french version of the Atom's Documentation and there is the first piece of the cake. I don't really know how to add this to the project so i'm open to any advice

Thanks you and hope you will like it!!


lee-dohm:

PR #316, opened 2017-01-19
Second part of the french translation

Bruneton:
Chapter 1: Installing Atom french translation


lee-dohm:

PR #317, opened 2017-01-19
Third part of the chapter 1 of the documentation in french

Bruneton:
Third part: atom basics in french


lee-dohm:

PR #318, opened 2017-01-19
Last part of chapter 1 french documentation

Bruneton:
last part of the chapter 1: summary in french

### DIFF
--- a/content/getting-started/sections/summary.md
+++ b/content/getting-started/sections/summary.md
@@ -1,8 +1,10 @@
 ---
-title: Summary
+titre: Sommaire
 ---
-### Summary
+### Sommaire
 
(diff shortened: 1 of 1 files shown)


lee-dohm:

Bruneton:
Hey,

First Thanks for your answer. That was our question because we weren't sure how this will be apllied to the project. 
For me, maybe, if we can add a button (like it's already used to choose between Mac, Windows or Linux) to select the langage (English or French) but I don't know how to do it.
Tell me if you think this idea can be add to the project.

Best Regards,

Sebastien

lee-dohm:
I think it can be added to the project and it is something that we would like to do. It would be a fairly extensive change though and we don't have people that can work on it right now 🙁 I'll see about writing an issue that describes, in general terms, what would need to be done.

PR #319, opened 2017-01-28
mention ATOM_DEV_RESOURCE_PATH when using custom location

dirk-thomas:
If the user has cloned Atom in the non-default location this will help to be aware that they have to set an environment variable in order for `atom --dev` work correctly.

### DIFF
--- a/content/behind-atom/sections/developing-node-modules.md
+++ b/content/behind-atom/sections/developing-node-modules.md
@@ -21,6 +21,10 @@ $ apm rebuild
 $ cd <em>WHERE YOU CLONED ATOM</em>
 $ npm link atom-keymap
 
+# If you have cloned Atom in a different location then ~/github/atom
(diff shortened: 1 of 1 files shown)


lee-dohm:
Thanks for the help :bow:

dirk-thomas:
That was quick - thank you!

50Wliu:
Two things:
* than not then
* This only works for Linux and macOS, a separate section should be added for Windows replacing `~/github/atom` with `%USERPROFILE%/github/atom` and `export` with `setx`.

dirk-thomas:
@50Wliu Thanks for the feedback.

@lee-dohm Thanks for taking care of it: 3d73d5898f7f1e2894f8be47cdfd4d1411808428

lee-dohm:
You're welcome @dirk-thomas 👍

PR #324, opened 2017-02-24
add waitsForPromise option parameter

UziTech:
This documents the options parameter for `waitsForPromise` at https://github.com/atom/atom/blob/master/spec/spec-helper.coffee#L237


lee-dohm:
Thanks for contributing 👍

PR #325, opened 2017-02-28
Add information on addKeystrokeResolver

lee-dohm:
Document the process where people can update their own keystroke resolution without creating a package.

/cc @nathansobo


nathansobo:
👍 Thanks for adding this.

Ben3eeE:
Wrote a review for this but I apparently forgot to submit it 😦 I wanted to mention that it might be good to add a check for `event.type` in the example because it runs for all types of events. Just something that got me when I was playing around with this 😄 


 ```coffee
 atom.keymaps.addKeystrokeResolver ({event}) ->
   if event.code is 'KeyG' and event.altKey and event.ctrlKey and event.type isnt 'keyup'
     return 'ctrl-@'
 ```

nathansobo:
@Ben3eeE Good point. You mind tossing something about that in?

PR #327, opened 2017-03-05
Add check for event.type in the addKeystrokeResolver example

Ben3eeE:
Refs https://github.com/atom/flight-manual.atom.io/pull/325#issuecomment-284122453


lee-dohm:
Thanks for cleaning up after me :wink:

PR #328, opened 2017-03-19
Add article stub for snapshots

lee-dohm:
/cc @nathansobo @as-cii to fill in the documentation for snapshots.

**[View Rendered Content](https://github.com/atom/flight-manual.atom.io/blob/snapshots/content/behind-atom/sections/how-atom-uses-chromium-snapshots.md)**


as-cii:
I added some brief information about how Atom uses snapshots, along with links to the relevant repositories and lines of code.

What do you think? 💭

nathansobo:
This looks great. One suggestion... maybe mention that people may need to modify the list of forbidden modules if they add a native dependency?

as-cii:
Thanks for the suggestion, @nathansobo. ⚡️ While I was editing this, I also clarified we strive to get as many things as possible into the snapshot, as opposed to excluding entire modules. 

If this looks good, please feel free to merge it after publishing the release. 👍

PR #329, opened 2017-03-19
Adds note for reloading changes.

### LINKED ISSUE #330: Word Count package tutorial should describe how to reload the window to see changes.
After making changes while following the [word count package tutorial](http://flight-manual.atom.io/hacking-atom/sections/package-word-count/), it's not obvious why changes to the packages code aren't automatically seen when toggling the package.

A small note should be added to describe that reloading the window will reload the package's code so the user may see their changes.

PR: #329

Orbyt:
Adds a simple note describing how to reload the window to see changes.

Fixes #330


lee-dohm:
Thanks for this!

PR #333, opened 2017-04-05
Remove warning about unpublishing before deleting

50Wliu:
It appears that this is no longer needed thanks to @jasonrudolph :zap:.
/cc @lee-dohm for review


lee-dohm:
Thanks for taking care of this @50Wliu :bow:

Ben3eeE:
@jasonrudolph 💯 ❤️

PR #334, opened 2017-04-06
Grammar fix (add an article)

bookofportals:
I added an article. Only "a" makes sense given the context, which is to check if the user is using the latest version.


50Wliu:
Thanks!

PR #340, opened 2017-04-27
Update installing-atom.md

bigraf:
Config settings for Windows 7 in a corporate environment.

After several hours and many failed attempts, it was the above proposal that worked for our team.

Setting both HTTP-PROXY and HTTPS-PROXY to the same value allowed Atom to access the internet.


lee-dohm:
It would be better to simply add the `http-proxy` part to the already existing instructions.

PR #342, opened 2017-05-08
Update find-and-replace.md

seranki:
Proposed adding this note at the end of line 15:
"Note: In the replacement part of a regex the syntax $1, $2, … $& must be used."
My "regex" experience is with languages that use the \1, \2 syntax, so the need to use Javascript syntax was not obvious to me.
Regards.

### DIFF
--- a/content/using-atom/sections/find-and-replace.md
+++ b/content/using-atom/sections/find-and-replace.md
@@ -12,7 +12,7 @@ If you launch either of those commands, you'll be greeted with the Find and Repl
 
 ![Find and replace text in the current file](../../images/find-replace-file.png "Find and replace text in the current file")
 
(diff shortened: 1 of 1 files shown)


lee-dohm:
Thanks for the contribution.

This would be better as a [Note admonition block](https://github.com/atom/flight-manual.atom.io/blob/master/CONTRIBUTING.md#admonition-blocks).

seranki:
I edited the pull request per your recommendation to make it a Note admonition block. I also moved it one para down where I think it fits better.
Since I am not a GitHub expert, please feel free to tell me how to correct any errors in how I did it.
Thanks.

lee-dohm:
Thanks for this!

seranki:
Thanks for your help. I will delete the branch now.

PR #344, opened 2017-05-13
Document lineHeight config key in editor options

michaelstepner:
lineHeight is a configuration option for the editor, but was not documented in the Flight Manual.


lee-dohm:
Thanks for this!

PR #349, opened 2017-06-14
GitHub package

simurai:
This adds a section about the GitHub package. Similar to https://github.atom.io but a little more detailed descriptions.


content/using-atom/sections/github-package.md
```diff
@@ -0,0 +1,120 @@
+---
+title: GitHub package
+---
+### GitHub package
+
+Atom's newest core package brings Git and GitHub integration right inside your editor!
```
50Wliu: Since the flight manual is a source of documentation, how about
> The github package brings Git and GitHub integration right inside Atom.

PR #351, opened 2017-06-19
Fix semantic versioning application

joepie91:
Semantic versioning covers the exposed interfaces of a package, not whether the developer perceives it as a 'bugfix' or otherwise. Therefore, the original explanation of `minor` vs. `patch` was incorrect.


joepie91:
Argh, typoed the title. One moment.

joepie91:
I can apparently not amend commit descriptions through the GitHub interface, so I'll reopen with a correctly-named commit.

PR #352, opened 2017-06-19
Fix semantic versioning explanation

joepie91:
(This is a fixed version of #351.)

Semantic versioning covers the exposed interfaces of a package, not whether the developer perceives it as a 'bugfix' or otherwise. Therefore, the original explanation of `minor` vs. `patch` was incorrect.


lee-dohm:
Since Atom packages don't truly have exposed interfaces, semver doesn't strictly apply. ([Atom packages that provide services](http://flight-manual.atom.io/behind-atom/sections/interacting-with-other-packages-via-services/) do have exposed interfaces, but those interfaces are versioned separately from the package itself.)

However, being a little more specific about things may help people. Thanks for the contribution!

PR #357, opened 2017-07-16
Fixed developer tools keybinds

chris579:
When merged this pull request will
- fix a keybind error for opening the developer tools in atom
- the keybind is `shift + ctrl + I` instead of `alt + ctrl + I`
- I'm unaware of mac but I'm quite sure that it should be shift too there


50Wliu:
Per the [Keybindings](https://github.com/atom/flight-manual.atom.io/blob/master/CONTRIBUTING.md#keybindings) section of the contributing guide, the Windows/Linux keybinding should be changed to Ctrl+Shift+I so that it stays alphabetical.

Other than that, this looks good.

chris579:
Reordered the keybinds + fixed the mac keybind too (I suppose that it is wrong too)

50Wliu:
Nope, the existing mac keybinding is correct :).
https://github.com/atom/atom/blob/3c2fc9494d480a82b3ddd427626b64b985db9000/keymaps/darwin.cson#L30

chris579:
Ok, well now it should be correct

PR #361, opened 2017-07-27
Clarified backspace and arrow key issues with components and input field

xentao:
Even after reading the original Flight Manual section, "Forcing Chromium's Native Keystroke Handling", I didn't make the connection that a bug with not being able to use backspace or arrow keys inside an included component (which has an input field) was related.   Clarified this in a tip.


lee-dohm:
Thanks for this! :+1:

PR #363, opened 2017-08-09
Fix the steps for developing node modules

Ben3eeE:
Some node modules, for example `atom-keymap`, `keyboard-layout` and `node-pathwatcher` requires `apm rebuild` to be run in the Atom directory on Windows. This PR changes the steps so `apm rebuild` is always run from the Atom directory.

/cc: @50Wliu

### DIFF
--- a/content/behind-atom/sections/developing-node-modules.md
+++ b/content/behind-atom/sections/developing-node-modules.md
@@ -14,11 +14,11 @@ $ git clone https://github.com/atom/atom-keymap.git
 $ cd atom-keymap
 $ npm install
 $ npm link
+$ cd <em>WHERE YOU CLONED ATOM</em>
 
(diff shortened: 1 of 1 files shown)


50Wliu:
Oh, does `apm rebuild` not actually have to be run in the node module's directory as well?

Ben3eeE:
I tested just rebuilding in the Atom directory for `atom-keymap` and that seemed to work. Would be good if you can test that out on a module as well just to verify. I have always done it twice before 😀

Ben3eeE:
@50Wliu Did you manage to get it working by only rebuilding in the `atom/atom` folder?

50Wliu:
Yeah, I think the only rebuild needed is atom/atom.

PR #374, opened 2017-09-18
[WIP] Add a grammar tutorial

50Wliu:
The goal of this PR is to create a mini tutorial series on how to create a grammar package.  Given the scope of grammar packages, it'll probably be much longer and more in-depth than the existing package + theme tutorials.  Similar to the package tutorials, it'll be split into different sections of varying difficulty.

@DamnedScholar I wanted to get your feedback on this.

### DIFF
--- a/content/hacking-atom/sections/creating-a-grammar.md
+++ b/content/hacking-atom/sections/creating-a-grammar.md
@@ -1,0 +1,186 @@
+---
+title: Creating a Grammar
+---
+### Creating a Grammar
+
+Grammars power Atom's syntax highlighting. Each grammar provides a set of _regex_ patterns which are assigned _scopes_. These scopes are then turned into the CSS classes that you can target in syntax themes!
+
+#### Getting Started
+
+Grammars depend heavily on regexes, and you should be comfortable with interpreting and writing regexes before continuing. Note that Atom uses the Oniguruma engine, which is very similar to the PCRE or Perl regex engines. Here are some resources to help you out:
+
+* http://www.regular-expressions.info/tutorial.html provides a comprehensive regex tutorial
+* http://www.rexegg.com/regex-quickstart.html contains a cheat sheet for various regex expressions
+* https://regex101.com/ or https://regexr.com/ allows live prototyping
+
+Grammar files are written in the [CSON](https://github.com/bevry/cson#what-is-cson) or [JSON](http://www.json.org/) format. Whichever one you decide to use is up to you, but this tutorial will be written in CSON.
+
+#### Create the Package
+
+<!-- TODO: This functionality doesn't exist yet... -->
+To get started, press <kbd class="platform-mac">Cmd+Shift+P</kbd><kbd class="platform-windows platform-linux">Ctrl+Shift+P</kbd> and start typing "Generate Grammar Package" to generate a new grammar package. Select "Generate Grammar Package," and you'll be asked for the path where your grammar package will be created. Let's call ours `language-flight-manual`.
+
+{{#tip}}
+
+**Tip:** Grammar packages should start with _language-_.
+
+{{/tip}}
+
+Open the `flight-manual.cson` file present in the `grammars` folder. This is the main file that we will be working with. Let's go over what each key means.
+
+`scopeName` is the root _scope_ of your package. This should generally describe what language your grammar package is highlighting; for example, `language-javascript`'s `scopeName` is `source.js` and `language-html`'s is `text.html.basic`. You can leave this as `scope.flight-manual` for now.
+
+`name` is the user-friendly name that is displayed in places like the status bar or the grammar selector. Again, this name should describe what the grammar package is highlighting. Rename it to `Flight Manual`.
+
+`fileTypes` is an array of filetypes that `language-flight-manual` should highlight. We're interested in highlighting the Flight Manual's Markdown files, so add the `md` extension to the list.
+
+`patterns` contains the array of regex patterns that will determine how the file is tokenized.
+
+#### Adding Patterns
+
+To start, let's add a basic pattern to tokenize the words `Flight Manual` whenever they show up. Your regex should look like `\bFlight Manual\b`. Here's what your `patterns` block should look like:
+```cson
+'patterns': [
+  {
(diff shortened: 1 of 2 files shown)


DamnedScholar:
Is there a newish templating engine or markup language that doesn't have enough of a community to have an Atom package yet? A language with maybe a couple dozen different entities would be perfect to have as a real-world example, and a single example that appears throughout the series and can also be observed in the reader's editor would be a great way to demonstrate how a language package comes together. Nobody should look at `language-java` as their first real example of a language package.

Are we trying to keep this focused and serious, or is there room at the end for some comments about fun things you can do in a language package outside of the `grammar

50Wliu:
I wanted this tutorial to feature a real-life example, but I can't think of any.  There's definitely room at the end for extra stuff.

content/hacking-atom/sections/creating-a-grammar.md
```diff
@@ -0,0 +1,186 @@
+---
+title: Creating a Grammar
+---
+### Creating a Grammar
+
+Grammars power Atom's syntax highlighting. Each grammar provides a set of _regex_ patterns which are assigned _scopes_. These scopes are then turned into the CSS classes that you can target in syntax themes!
```
DamnedScholar: For the absolute newbies, maybe this should be "*regex* (regular expression) patterns". The only mention of the full name is two paragraphs down in a URL.

content/hacking-atom/sections/creating-a-grammar.md
```diff
@@ -0,0 +1,186 @@
...
+
+#### Getting Started
+
+Grammars depend heavily on regexes, and you should be comfortable with interpreting and writing regexes before continuing. Note that Atom uses the Oniguruma engine, which is very similar to the PCRE or Perl regex engines. Here are some resources to help you out:
+
+* http://www.regular-expressions.info/tutorial.html provides a comprehensive regex tutorial
+* http://www.rexegg.com/regex-quickstart.html contains a cheat sheet for various regex expressions
+* https://regex101.com/ or https://regexr.com/ allows live prototyping
+
+Grammar files are written in the [CSON](https://github.com/bevry/cson#what-is-cson) or [JSON](http://www.json.org/) format. Whichever one you decide to use is up to you, but this tutorial will be written in CSON.
```
DamnedScholar: I'm going to be very pedantic here, but I think that this would read better as, "Grammar files are written in either CSON or JSON format."

content/hacking-atom/sections/creating-a-grammar.md
danPadric: If you hope to provide for newbies, you will have to have text that starts at first in terms even more basic than scopes and css classes. Maybe best to show first what it can do and how you can do it, without the technicalities. Get the 'client' interested first and then follow up by terminology IMHO.

DamnedScholar: I don't think it needs to target people who don't know what CSS classes are. Obviously, it *could*, but how many people at that level of familiarity are going to be making a language package?

danPadric: > how many people at that level of familiarity are going to be making a language package?
At least one... **_me_**. As I am not so special, there should be more people like that.

Following your logic, the people creating a grammar should then be people who knows regular expressions or regex or RE. There should not even be a web link to what it is and what it does.

\- thinking aloud - the double escaping in the regular expression patterns was very much a surprise.

DamnedScholar: There are plenty of tutorial resources for CSS that handle it much better than an off-hand explanation in a grammar tutorial would, and I would have anyone who doesn't know CSS read a couple of them before starting. If you need, I can direct them to that. Note that I also didn't recommend that we instruct people on how to use regular expressions, only that we should explain what "regex" is short for. This is because there are plenty of regex tutorials. We could link to both kinds of tutorials, but we shouldn't devote paragraphs to them.

> - thinking aloud - the double escaping in the regular expression patterns was very much a surprise.

This is a unique quirk, based on the fact that CSON has to be parsed into JSON, and isn't applicable under the circumstance where the package author chooses to write in JSON. It definitely deserves a line or two, because it's not expected behavior, but it would never be part of any standard regex tutorial.

50Wliu:
If anyone has suggestions please feel free to create a PR targeted at this branch.  While I'd like to finish this it depends on motivation.

50Wliu:
Now available at http://flight-manual.atom.io/hacking-atom/sections/creating-a-grammar/.  Please feel free to create PRs to improve it!

udnaan:
One of the issues that one faces writing a language grammar that targets a non-trivial language for atom is that while it has been easy enough to look at existing language packages and come up with a json/cson file; it is much harder to create something that highlights code blocks. 

There is hardcoded support for matching brackets ({} () etc), but what if you have a language with custom block start/end/else tags.

A simple example is XML (again the support is hardcoded in atom) or a custom templating language with multiple levels of blocks.

DamnedScholar:
> There is hardcoded support for matching brackets ({} () etc), but what if you have a language with custom block start/end/else tags.

That's what `begin` and `end` are for. Inside the code block pattern, you can include a subset or even the whole set of patterns for that particular language using `include: "$self"` as discussed in [the tutorial](https://flight-manual.atom.io/hacking-atom/sections/creating-a-grammar/#repositories-and-the-include-keyword-or-how-to-avoid-duplication).

If you would like help designing your grammar, please make a thread on [the Atom forums](https://discuss.atom.io) and I'd be happy to tell you what you need to do.

udnaan:
It was only a comment on what would be nice to have in a tutorial about language extension/grammar.
I have personally found that writing such tutorial can help figure out short coming in the API.

DamnedScholar:
The tutorial also shows how to include external patterns inside custom blocks, at the end of that section I linked.

udnaan:
I haven't gone through the tutorial yet but if you think it already covers the pattern I mentioned, than that's great. 
Thanks for adding it!

50Wliu:
Yeah, the tutorial's still a bit rough but I think it somewhat covers what you're asking :).

PR #388, opened 2017-10-26
Update windows shortcut to developer tool

massivelines:
The windows shortcut to open the developer tool was 
Alt+Ctrl+I
but needs to be
Ctrl+Shift+I


Ben3eeE:
Thanks for contributing :bowing_man:

PR #389, opened 2017-10-26
Update the windows reload shortcut

massivelines:
The shortcut for windows, window reload was
Alt+Ctrl+R
but needs to be updated to
Ctrl+Shift+F5


Ben3eeE:
Thanks for contributing :bowing_man:

Issue #390, opened 2017-10-26
Improve the steps for developing node modules


Ben3eeE:
http://flight-manual.atom.io/behind-atom/sections/developing-node-modules/

* The steps feel uncertain because they end with `Should work!`
  * What can we do if it doesn't work?
  * How can we tell if it did or did not work?
  * Is there some command to check for links available like `apm links`?
* There are no steps for how to unlink something when you are done with testing the module
* There is no mention for cases when this gets crazy: https://github.com/atom/atom/pull/13943#issuecomment-324475662
  * Do we need to link pathwatcher to atom/atom when linked to atom/text-buffer in this case?
  * Why?
  * Maybe an example of testing some change like this out?

/cc: @ungb

Ben3eeE:
/cc: @50Wliu You might have some input here.

50Wliu:
> Is there some command to check for links available like apm links?

Nope.  However, I believe running `npm ls` will show any links in the dependency tree if they exist.

> There are no steps for how to unlink something when you are done with testing the module

This frustrates me; as far as I can tell the only option you have from the CLI is to completely uninstall the module, which also removes it from `package.json`.  Then add it back.

PR #391, opened 2017-10-27
Fix Node module rebuilding after the initial steps

50Wliu:
/cc @Ben3eeE we forgot to update the second command-line section :P.

Refs #390


Ben3eeE:
oops :sweat_smile:

PR #393, opened 2017-11-02
Add note on how to customize keybindings in editing and deleting section of atom

ungb:
Creating this PR to address https://github.com/atom/atom/issues/4026 

where there's confusion about what keybinding does what and make it more clear in the docs that people can change them to suit their needs.

/cc @nathansobo @rsese

### DIFF
--- a/content/using-atom/sections/editing-and-deleting-text.md
+++ b/content/using-atom/sections/editing-and-deleting-text.md
@@ -39,6 +39,14 @@ You can also delete or cut text out of your buffer with some shortcuts. Be ruthl
 
 {{/mac}}
 
+
+{{#note}}
+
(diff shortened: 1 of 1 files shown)


nathansobo:
Would it be worthwhile to mention word boundaries explicitly since this is a common hangup for people?

rsese:
I think the general note is helpful but since it's a high level note, it seems a little out of place specifically in the "Deleting and Cutting" section?

> Would it be worthwhile to mention word boundaries explicitly since this is a common hangup for people?

I looked at my notes and it seems like we talked about adding docs specifically to address https://github.com/atom/atom/issues/4026?  Would we want to keep the high level note and add more details to address https://github.com/atom/atom/issues/4026, or remove the note and just add a small paragraph to address the issue?

nathansobo:
I'd like to find some place to put an explicit mention of word-based movement and deletion boundaries and give an example of how people might customize them.

ungb:
closing in favor of https://github.com/atom/flight-manual.atom.io/pull/426

PR #397, opened 2017-11-13
Just updating screenshot for different wording

ItsPugle:
Simply updating the screenshot, as in the current Atom version, the wording is slightly changed.


thomasjo:
Thanks for helping out! I noticed you're linking to an external image. You're going to have to change the actual image stored in the repo instead of changing the URI.

The screenshot image you'll want to change is the one in the original URI, which you can find at [`content/using-atom/images/menubar.png`](https://github.com/atom/flight-manual.atom.io/blob/master/content/using-atom/images/menubar.png).

ItsPugle:
I spent ages looking for that, but couldn't find it! Guess I wasn't looking hard enough 😄  I've gone ahead and (hopefully) fixed this, now!

50Wliu:
Can you keep `Stylesheet...` highlighted like it was previously?

ItsPugle:
Done, @50Wliu! How does it look now?

thomasjo:
@ItsPugle Thanks again for helping out, we really appreciate it :bowing_man:

ItsPugle:
Woot woot! Cheers!

PR #403, opened 2017-11-28
Replaced reload:window keybind for linux and win32 with the correct one

nineway:
Replaced reload:window keybind for linux and win32 with the correct one in the debugging section.

Sorry for all the commits, had some trouble with contributing to another branch.


50Wliu:
`f5` should be capitalized.  Don't worry about the commits :).

Issue #380, opened 2017-10-09
Update Check for Errors in the Developer Tools


Ben3eeE:
This section http://flight-manual.atom.io/hacking-atom/sections/debugging/#check-for-errors-in-the-developer-tools could use an update. It would be good to rewrite this and describe the following things:

1. That not all errors are logged through the notifications and that error messages are sometimes only in the developer tools console
1. How to open the developer tools in Atom
1. How to open the console in the developer tools
1. How to check if there is an error or warning in the developer tools console

PR #404, opened 2017-11-28
Updated check for errors in the debugging section

nineway:
This PR adds more detail about how to check for errors, how to open developer tools console and that not all errors are logged by the red alert notification.

#380

(Sorry for all the commits)


rsese:
Ugh, I lost track of https://github.com/atom/flight-manual.atom.io/issues/380, I was supposed to open a PR for that one 😭  Thanks for taking this on @nineway :bow:

Here's what I had:

```diff
diff --git a/content/hacking-atom/sections/debugging.md b/content/hacking-atom/sections/debugging.md
index 3e50e6a..790d7d2 100644
--- a/content/hacking-atom/sections/debugging.md
+++ b/content/hacking-atom/sections/debugging.md
@@ -208,12 +208,16 @@ When an unexpected error occurs Atom, you will normally see a red notification w
 
 If you can reproduce the error, use this approach to get the full stack trace and [report the issue](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#submitting-issues).
 
+Not all errors are logged with a notification however so if you suspect you're experiencing an error but there's no notification, you can also look for errors in the developer tools Console tab.  To access the Console tab, press <kbd class="platform-mac">Alt-Cmd-I</kbd><kbd class="platform-windows platform-linux">Ctrl-Shift-I</kbd> to open developer tools and then click the Console tab:
+
+![DevTools Error](../../images/devtools-error.png)
+
+If there are multiple errors, you can scroll down to the bottom of the panel to see the most recent error.  Or while reproducing an error, you can right click in the Console tab panel, select `Clear console` to remove all Console output, and then reproduce the error to see what errors are logged to the Console tab.
+
 {{#note}}
 
 **Note:** When running in Dev Mode, the developer tools are automatically shown with the error logged in the Console tab.
 
-![DevTools Error](../../images/devtools-error.png)
-
 {{/note}}
 
 #### Find Crash Logs
```

I think our first paragraphs are basically equivalent.  Then I also moved the image in the note:



I moved it out of the note and put it right after the first paragraph where we explain how to access devtools and the console tab since I thought that's helpful for the explanation.  Then since we just showed the image, I think the note still makes sense on its own without the image.  What do you think about adding that change?

I also added a second paragraph with a bit more detail about checking for an error - if you think it's useful I think we can add it this PR but if not I can propose it in a followup.

nineway:
Yeah, I think that your explanation is better so I altered it as you said and the image didn't say much so I also deleted it. However, we don't answer question 4 from the original issue now "How to check if there is an error or warning in the developer tools console". Do you think it's necessary to add?

rsese:
> However, we don't answer question 4 from the original issue now "How to check if there is an error or warning in the developer tools console". Do you think it's necessary to add?

Hmmm, I think we cover checking an error ok but you're right, we didn't add anything about warnings.  Since what we have is an improvement, I think this looks good as is and we can talk about warnings in a separate PR maybe.  If that sounds good to you, maybe we can get a final set of 👓  from @lee-dohm on these changes?

lee-dohm:
Warnings look very similar to errors, so I don't think we really need to call them out separately.

Since the errors reported here aren't due to this PR, I'm going to go ahead and merge this. Thanks very much for the help! :tada:

rsese:
Thanks for the 👓 @lee-dohm and thanks for your contributions @nineway! :bow:

PR #407, opened 2017-11-29
Whenever "open source" is used as an adjective before a noun, it shou…

JosephTLyons:
…ld be hyphenated.

https://www.englishforums.com/English/OpenSourceOrOpenSource/bzpcd/post.htm

All other instances of "open source" are adjectives after the noun, and don't require a hyphen.


lee-dohm:
Would you mind adding something to the `CONTRIBUTING.md` for this as part of this change?

JosephTLyons:
I don't mind at all, however, I looked through that file and wasn't sure what exactly you were looking for in terms of a new addition to it.  I'll gladly add to it though, just would like some guidance first.

lee-dohm:
I think it would fit rather well into the [Language section](https://github.com/atom/flight-manual.atom.io/blob/master/CONTRIBUTING.md#language), following the **DO** and **DO NOT** format.

JosephTLyons:
I added a DO bullet point.  Let me know if you'd like it revised.

lee-dohm:
Thanks very much!

PR #415, opened 2017-12-19
I guess the scopes are defined inverse than in the doc

rriemann:


50Wliu:
The way you define the scopes doesn't matter, and I believe that settings-view will actually reorganize the scopes to be how they currently are listed in the Flight Manual.

Thank you for your contributions though!

PR #416, opened 2017-12-20
Add instructions for installing Atom using Linux package repos

daviwil:
I've updated the Linux installation instructions to add steps for configuring package repositories for distributions using the .deb and .rpm formats.  Would love for @ungb to double-check the steps before we merge this just to make sure everything works correctly :)

### DIFF
--- a/content/getting-started/sections/installing-atom.md
+++ b/content/getting-started/sections/installing-atom.md
@@ -70,13 +70,30 @@ With Atom open, click on `File > Settings`, and then the `System` tab on the lef
 
 #### Installing Atom on Linux
 
-To install Atom on Linux, you can download a [Debian package](https://atom.io/download/deb) or [RPM package](https://atom.io/download/rpm) either from the [main Atom website](https://atom.io) or from the [Atom project releases page][releases]. These packages do not currently have auto-update features, so when you would like to upgrade to a new release of Atom, you will have to repeat this installation process.
-
-[releases]: https://github.com/atom/atom/releases/latest
+You can install Atom on Linux using your distribution's package manager by configuring it to use one of our official package repositories. This will also enable you to update Atom when new releases are published.
 
 ##### Debian and Ubuntu (deb/apt)
 
-To install Atom on Debian, Ubuntu, or related systems:
+To install Atom on Debian, Ubuntu, or related distributions, add our official
+package repository to your system by running the following commands:
+
+``` command-line
+$ curl -L https://packagecloud.io/AtomEditor/atom/gpgkey | sudo apt-key add -
+$ sudo sh -c 'echo "deb [arch=amd64] https://packagecloud.io/AtomEditor/atom/any/ any main" > /etc/apt/sources.list.d/atom.list'
+$ sudo apt-get update
+```
+
+You can now install Atom using `apt-get` (or `apt` on Ubuntu):
+
+``` command-line
+# Install Atom
+$ sudo apt-get install atom
+
+# Install Atom Beta
+$ sudo apt-get install atom-beta
+```
+
+Alternatively, you can download the [Atom .deb package](https://atom.io/download/deb) and install it directly:
 
 ``` command-line
 # Install Atom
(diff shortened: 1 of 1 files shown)


content/getting-started/sections/installing-atom.md
```diff
@@ -86,25 +105,55 @@ $ sudo dpkg -i atom-amd64.deb
 $ sudo apt-get -f install
 ```
 
-##### RedHat and CentOS (yum)
+##### RedHat and CentOS (yum) or Fedora (dnf)
```
daviwil: Is it OK to merge these sections since the instructions are very similar?

content/getting-started/sections/installing-atom.md
```diff
@@ -70,13 +70,32 @@ With Atom open, click on `File > Settings`, and then the `System` tab on the lef
...
+
+```
+# Install Atom
+sudo apt-get install atom
+
+# Install Atom Beta
+sudo apt-get install atom-beta
+```
+
+You can also download the Atom .deb package and install it directly:
```
daviwil: Should I say "Alternatively, you can..."?

content/getting-started/sections/installing-atom.md
```diff
@@ -70,13 +70,32 @@ With Atom open, click on `File > Settings`, and then the `System` tab on the lef
...
 [releases]: https://github.com/atom/atom/releases/latest
 
 ##### Debian and Ubuntu (deb/apt)
 
-To install Atom on Debian, Ubuntu, or related systems:
+To install Atom on Debian, Ubuntu, or related distributions, add our official
+package repository to your system by running the following commands:
+
+```
+curl -L https://packagecloud.io/AtomEditor/atom/gpgkey | sudo apt-key add -
```
daviwil: There seems to be a convention of adding `$` in front of command lines in this document but I didn't use it in these blocks so it's easier to copy/paste into a terminal.  What do you think?

lee-dohm: Command line sections should be done with:

``````
``` command-line
$ command <em>replaceable parameter</em>
```
``````

The `$` gets converted to `C:\>` on Windows and the prompt isn't selectable. It's magic :sparkles:

File: content/getting-started/sections/installing-atom.md (hunk 2 above)
Comment (lee-dohm) [reply content/getting-started/sections/installing-atom.md]: Yes, that works :+1:

File: content/getting-started/sections/installing-atom.md (hunk 1 above)
Comment (lee-dohm) [reply content/getting-started/sections/installing-atom.md]: Yes, it appears to be clear enough.

File: content/getting-started/sections/installing-atom.md [hunk 4]
```diff
@@ -70,13 +70,32 @@ With Atom open, click on `File > Settings`, and then the `System` tab on the lef
...
+sudo sh -c 'echo "deb [arch=amd64] deb https://packagecloud.io/AtomEditor/atom/any/ any main" > /etc/apt/sources.list.d/atom.list'
+sudo apt-get update
```
50Wliu: I believe it's now recommended to use `apt` as a more beginner-friendly alternative, at least on Ubuntu.  Same for `apt install`.

daviwil: Thanks for pointing that out!  I think `apt` may only be available on Ubuntu variants and not Debian, so it may be better to use `apt-get` since it's the most universal command.  I'll mention that `apt` can also be used

daviwil:
Updated the file to fix the command blocks and fixed an issue with the Debian repo configuration thanks to @ungb's help!

content/getting-started/sections/installing-atom.md
```diff
@@ -70,13 +70,32 @@ With Atom open, click on `File > Settings`, and then the `System` tab on the lef
 
 #### Installing Atom on Linux
 
-To install Atom on Linux, you can download a [Debian package](https://atom.io/download/deb) or [RPM package](https://atom.io/download/rpm) either from the [main Atom website](https://atom.io) or from the [Atom project releases page][releases]. These packages do not currently have auto-update features, so when you would like to upgrade to a new release of Atom, you will have to repeat this installation process.
+To install Atom on Linux, you can download a [Debian package](https://atom.io/download/deb) or [RPM package](https://atom.io/download/rpm) either from the [main Atom website](https://atom.io) or from the [Atom project releases page][releases]. If you'd like Atom to be updated when new releases are published, you can configure an official package repository for your distribution.
```
ungb: Hmm I find it weird that we tell them they can download from the main site but now that the instructions are more towards installing using the package repository.  I wonder if we should remove the section about being able to download the builds or maybe have a separate section for manually installing.

daviwil: Yeah, I thought about that too.  Since I provide the alternative of installing the packages individually (with links to the download link in each distro section), I can remove these links from the top.  How about this wording instead:

> You can install Atom on Linux using your distribution's package manager by configuring it to use one of our official package repositories.  This will also allow you to update Atom when new releases are published.

ungb: Sounds good!  Should we include a section about how to manually install too?  I guess we shouldn't since this is way better.

daviwil: The manual install bits are still there, check out these sections, just search for the word "Alternatively" and you'll see the manual steps for the 3 package managers

daviwil:
Thanks for all the feedback!  Merging this now.  If you think of any other changes I should make I'll be happy to open another PR.

PR #417, opened 2017-12-21
Force TravisCI to use Node.js 7 to avoid node-sass compatibility error

daviwil:
This change forces TravisCI to use Node.js 7 instead of the system-wide
install (Node.js 9) to avoid version compatibility issues with the
node-sass package.  node-sass has builds against specific Node.js versions
and can't be used with anything newer than Node.js 7 unless we upgrade
our node-sass dependency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ cache:
 sudo: false
 git:
   depth: 10
+before_install:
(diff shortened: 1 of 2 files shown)


lee-dohm:
Thanks for taking care of this :+1:

PR #420, opened 2017-12-24
Add windows-build-tools to build instructions for Windows

ericcornelissen:
As I suggested in https://github.com/atom/atom/issues/15368, this adds [windows-build-tools](https://www.npmjs.com/package/windows-build-tools) to the build requirements as an option for Visual Studio. That title is also changed to C++ build tools, as it no longer only lists products directly related to Visual Studio.

### DIFF
--- a/content/hacking-atom/sections/hacking-on-atom-core.md
+++ b/content/hacking-atom/sections/hacking-on-atom-core.md
@@ -68,10 +68,11 @@ In order to build Atom from source, you need to have a number of other requireme
   * The `python.exe` must be available at `%SystemDrive%\Python27\python.exe`. If it is installed elsewhere create a symbolic link to the directory containing the `python.exe` using: `mklink /d %SystemDrive%\Python27 D:\elsewhere\Python27`
 * 7zip (7z.exe available from the command line) - for creating distribution zip files
 * C++ build tools, either:
-  * [windows-build-tools](https://www.npmjs.com/package/windows-build-tools) (Also installs Python 2.7.x)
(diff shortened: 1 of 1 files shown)


lee-dohm:
@damieng @daviwil What do you think about this addition? I don't know how stable/supported this npm package is, so I can't evaluate whether we would want to support this.

daviwil:
Thumbs up for `windows-build-tools`, it installs all the necessary C++ toolchain with zero headache.  It also installs Python but for some reason our build script doesn't pick it up (unless you have to reboot before it shows up in the PATH).  I'd say it's a good recommendation.

ericcornelissen:
>  It also installs Python but for some reason our build script doesn't pick it up (unless you have to reboot before it shows up in the PATH)

No it doesn't, see [this](https://github.com/felixrieseberg/windows-build-tools/issues/18) issue (even though I thought it did 😅). So I guess the `(Also installs Python 2.7.x)` should be removed or changed for clarity...

daviwil:
It's possible we could add a check to the Windows build script to look for Python under the `%HOME\.windows-build-tools`, then the prerequisite installation step will be much easier for Windows users.

damieng:
I've not had any headache with the C++ Build Tools being installed directly and I'm reluctant to support a third way to get the C++ deps on the system especially as it involves a 3rd party package.

lee-dohm:
Updated the PR with extra language since it isn't an official method of installation. Also placed it at the bottom of the list since we can't recommend it because again, it isn't official.

lee-dohm:
Thanks for the contribution :+1:

PR #422, opened 2017-12-26
Use local paths whenever possible and place favicon to root

Ingramz:
This pull request improves upon #421 and does two noteworthy things:

1) Removes `https://flight-manual.atom.io` from links within articles. This makes such links protocol agnostic and when viewing a locally compiled version, you'll be not redirected out of your local copy.

2) Moves `favicon.ico` to the root of domain.
The gulp task for copying the favicon already copied the icon file to the root of output, however the automatic prune functionality from nanoc deleted the file quickly after. Adding the file to nanoc exclusion list solves that.

   In templates it is still beneficial to reference `/favicon.ico` as the icon of choice in case the browser does not look for it automatically. I moved the relevant `<link>` tag closer to others, which simplifies things a bit. The path to favicon is now also protocol agnostic.


lee-dohm:
Looks good to me, thanks!

PR #427, opened 2018-01-12
Bug with apm config ssl

SJhne92:
apm config ssl just adds: strict.sl=false


50Wliu:
Please make sure you're typing the command correctly as `apm` just copies whatever you give it to the config.

PR #428, opened 2018-01-13
Brevity of Introduction

LeoSebastian:
I was thrilled by Atom Basics page but felt that the summary page offered no valuable information


lee-dohm:
Thanks for the suggestion.

We would rather enhance the summary if it isn't useful as-is than just removing it. Especially since every other chapter has a summary.

PR #433, opened 2018-01-18
Update example output of "html" snippet

thutterer:


50Wliu:
This is still outdated - in upcoming versions of Atom `lang` and `dir` attributes have been added to the html tag.
https://github.com/atom/language-html/blob/5b71da7da1807e93659b45690a7af75782cc5cd7/snippets/language-html.cson#L164

thutterer:
Ah, thanks for looking ahead @50Wliu ... But my PR uses the current version of Atom (1.23.3) which still uses an older release of language-html package (0.48.5) so I think it makes sense to merge this, as it is what the vast majority of users get with the snippet.
I can keep an eye open for when this changes and create another PR. I use the snippet often enough :smiley:

50Wliu:
I think to prevent having to update this snippet twice in a short period of time, I'll just wait until the new snippet is on Atom stable and then ask you to change the docs and merge :).  Hope you don't mind about the delay.

thutterer:
> Hope you don't mind about the delay.

Not at all! Let's do it like that :+1:

PR #436, opened 2018-01-27
Add a list of useful resources

Aerijo:
The first change is to add a link to the Oniguruma docs in the section about regular expressions. I feel this is important, to definitively point people to the right documentation.

The second change is to add a new section with links to other guides / useful information. My own guide was suggested to be included by @50Wliu, but I felt the others are all excellent as well (I know I depended on them). As I say in the short description, some links are present earlier in the guide. However, I feel it's good to have them all grouped together as well, to centralise everything.

I'm open to any suggested changes. I chose to use linked text instead of the raw links (in most cases) as I feel it helps explain the purpose of the link.

### DIFF
--- a/content/hacking-atom/sections/creating-a-grammar.md
+++ b/content/hacking-atom/sections/creating-a-grammar.md
@@ -183,7 +183,7 @@ However, including `source.gfm` has led to another problem: note blocks still do
 ]
 ```
 
-#### Where to go from here
+#### Where to Go from Here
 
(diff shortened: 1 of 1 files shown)


Aerijo:
Seems Travis fails because of this link https://wiki.gnome.org/Projects/Libsecret, which is down right now.

lee-dohm:
This is awesome, @Aerijo ... thanks!

PR #438, opened 2018-02-06
Reword ATOM_DEV_RESOURCE_PATH documentation

### LINKED ISSUE #435: Confusing instruction for ATOM_DEV_RESOURCE_PATH for Windows.
In [Hacking on Atom Core for Windows](https://flight-manual.atom.io/hacking-atom/sections/hacking-on-atom-core/#platform-windows) I found the following instruction.

> But first, you have to set the ATOM_DEV_RESOURCE_PATH environment variable. It defaults to %USERPROFILE%\github\atom which then expects the atom/atom repository to be cloned into %ATOM_DEV_RESOURCE_PATH%.

The Flight Manual tells the user that they "have to" set the environment variable, but then immediately spins off to describe the default setting, with no mention of what the user _should_ set the variable to. I read it over 20 times before I finally managed to parse out what I _think_ it is trying to say? _If_ I am correct, then this attempting to instruct me to _either_ clone the repo to `%USERPROFILE%\github\atom` _or_ set the environment variable to point to the root of my Atom repo. This was _way_ too confusing to get to.

And upon further study of the guide, it's starting to look like as a package developer, setting this variable at all will be quite problematic to begin with (quite the opposite of "have to"), because it will make it very difficult to run packages from source in the Release version of Atom.

50Wliu:
@skylize is this better?

Fixes #435

### DIFF
--- a/content/hacking-atom/sections/hacking-on-atom-core.md
+++ b/content/hacking-atom/sections/hacking-on-atom-core.md
@@ -26,7 +26,7 @@ $ script/bootstrap
 
 #### Running in Development Mode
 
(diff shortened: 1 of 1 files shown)


lee-dohm:
Thanks @50Wliu :grinning:

PR #445, opened 2018-02-27
Atom's number of packages greatly increased now

thadguidry:
According to Search Box here https://atom.io/packages there are 7,362 packages today.


50Wliu:
Thanks for the contribution!  However, when we say "by default", we mean the packages that are immediately available in Atom without having to install anything else.  For that reason, 90 is still the correct number.

thadguidry:
@50Wliu Ah, understood...then can I suggest that it is probably even better to expand and say that... "has more than 90 default packages and more than 7300 available packages".  I wondered why only a small 90, and then when I saw much more available packages I was encouraged even more.  This page is highly visible from search engines, so I'd recommend to expand with my suggestion to help advertising. :)

50Wliu:
@thadguidry Great idea!  I think we'd accept a PR with something similar to that.

PR #447, opened 2018-02-28
correct mac keybingings

MichaelDimmitt:
Sourced from https://gist.github.com/chrissimpkins/5bf5686bae86b8129bee
and tested locally on two mac computers with different environments.

### DIFF
--- a/content/using-atom/sections/folding.md
+++ b/content/using-atom/sections/folding.md
@@ -5,10 +5,12 @@ title: Folding
 
 If you want to see an overview of the structure of the code file you're working on, folding can be a helpful tool. Folding hides blocks of code such as functions or looping blocks in order to simplify what is on your screen.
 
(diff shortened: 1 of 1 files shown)


MichaelDimmitt:
The keybindings are showing but show up small on the readme.
Made a few attempts, not sure how to make the new keybindings show with a larger font.

50Wliu:
Thanks for the contribution!  However, the existing keybindings are correct and use the modifier names rather than symbols for consistency with other platforms.

50Wliu:
This is a blurry image, but it seems to be the default Magic Keyboard that most macOS computers ship with:
magic-keyboard

The standard shortcut identifiers are present, and hence I would argue that most macOS users are familiar with cmd, alt, shift, etc.

In addition, the usage of platform-agnostic keybindings means that the keybindings presented in the Flight Manual is a more direct mapping to Atom's internal keybinding system than the macOS symbols are.

PR #449, opened 2018-03-11
Fixes issue #10138

kejn:
Please see the errors described in:
* https://github.com/atom/atom/issues/10138
* https://github.com/atom/atom/issues/15155

When running bootstrap script there occurred an error:

    ENOENT: no such file or directory, rename '/some/path/node_modules/module' -> /some/node_modules/.module.delete

Installing the dependencies described in the 'Building' section solves the problem and the bootstrap script finishes without errors.

### DIFF
--- a/content/hacking-atom/sections/hacking-on-atom-core.md
+++ b/content/hacking-atom/sections/hacking-on-atom-core.md
@@ -24,6 +24,13 @@ $ cd <em>where-you-cloned-atom</em>
 $ script/bootstrap
 ```
 
+###### Ubuntu / Debian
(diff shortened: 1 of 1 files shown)


lee-dohm:
Thanks for the contribution.

Because this PR doesn't fix any of the issues mentioned here, I'm not going to accept it as-is. Additionally, rather than saying here in the PR that installing the dependencies described in the build section will fix the problem ... the text added to the Flight Manual should say that, so that we don't have to edit the list in two places.

Please do submit another PR with the above in mind.

PR #453, opened 2018-03-21
Add additional information about regex to find-and-replace.md

### LINKED ISSUE #452: Regex usage (Find and Replace) poorly explained
## Why I think the information provided is inadequate:

Recently, I tried to use the _Regex_ option in _Find and Replace_ for the first time. I did not know how to use it, so I referred to the Flight Manual. The only information I got from there is [this note](https://flight-manual.atom.io/using-atom/sections/find-and-replace/):

> Note: With a regular expression search, the replacement syntax to refer back to search groups is $1, $2, … $&

I understood that you can use `.*` to include any arbitrary string of characters into your _search expression_ and then use `$1` to refer back to it in the _replace expression._ So I did this:

> Find: `text1 .* text2`
> Replace: `text3 $1 text2`

And I wondered why the text covered by “.\*” was not preserved but instead replaced by “$1”. (I now know that I have to enclose “.\*” in parentheses like this: `text1 (.*) text2`)

I did web searches using the terms “atom regex” and “atom regular expression”; they resulted in absolutely nothing that could help with my problem.

Later I was scrolling through the Wikipedia article on _regular expressions_ (which is very technical and not meant to be a tutorial). I stumbled across [this table](https://en.wikipedia.org/wiki/Regular_expression#POSIX_basic_and_extended) and read the line:

Metacharacter | Description
-|-
**`( )`** | Defines a marked subexpression. The string matched within the parentheses can be recalled later (see the next entry, `\n`). A marked subexpression is also called a block or capturing group. **BRE mode requires `\( \)`.**

That helped me solving my problem from above.

I still don’t know which other metacharacters you can use with _Find and Replace._

## My suggestions for improvement:

* Say which standard of regex syntax is being used so you can look it up yourself
* Link to an external page where you can learn about how to use it
* Include a table with all metacharacters that can be used (like the one from Wikipedia)

craaaa:
Fixes #452

- Specified that Atom uses JavaScript regex syntax
- Added a link to JavaScript's documentation

### DIFF
--- a/content/using-atom/sections/find-and-replace.md
+++ b/content/using-atom/sections/find-and-replace.md
@@ -18,7 +18,9 @@ If you type a string in the replacement text box, you can replace matches with a
 
 {{#note}}
 
-**Note:** With a regular expression search, the replacement syntax to refer back to search groups is  $1, $2, … $& 
+**Note:** Atom uses JavaScript regular expressions to perform regular expression searches.
+
+When doing a regular expression search, the replacement syntax to refer back to search groups is  $1, $2, … $&. Refer to JavaScript's [guide to regular expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) to learn more about regular expression syntax you can use in Atom.
 
 {{/note}}
 


craaaa:
@50Wliu would appreciate if you could take a look at this! Thanks :)

lee-dohm:
Thanks for the contribution 👍

PR #462, opened 2018-04-19
Document search feature

daiyi:
This was oddly unintuitive to me. I followed the wikipedia link and read all about globs which was deceivingly unhelpful in helping me figure out how to search for multiple globs at a time, so I thought I'd add this tip for posterity (along with the other super helpful tips in this section!)


lee-dohm:
Thanks for helping out :bow:

PR #463, opened 2018-04-23
added hint that package activation may be async

jneuendorf:
I have been searching for an information on this topic. All I've found was an old locked issue and this:

https://github.com/atom/atom/blob/d6a3a604695df62a7b813bd229316042a501aee5/src/package-manager.js#L678-L704

So I hope I didn't misunderstand the code concerning package activation and this PR helps people find that info more quickly .


lee-dohm:
Thanks for the contribution :bow:

PR #464, opened 2018-04-24
Update publishing.md

theclearcoder:
Include a comment about using a Personal Access Token when 2FA is on.


rsese:
Thanks @theclearcoder :bow:

PR #466, opened 2018-05-04
Break commands in two lines at linux install instructions

danieldeveloper001:
## Description 

When the commands are put together in the same line at bash, the admin password and curl prompts are mixed at bash, which may lead to a confusing experience for the Atom's users during the installation process (the user may not see at first the sudo prompt and keep waiting for the curl result). Thanks for considering this little change.

image

## Environment

- Operating System: Ubuntu Linux 17.10

## Notes
 
Sorry about the branch name, I've just made the change though GitHub's interface.

### DIFF
--- a/content/getting-started/sections/installing-atom.md
+++ b/content/getting-started/sections/installing-atom.md
@@ -78,7 +78,8 @@ To install Atom on Debian, Ubuntu, or related distributions, add our official
 package repository to your system by running the following commands:
 
 ``` command-line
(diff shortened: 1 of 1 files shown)


danieldeveloper001:
Please, just ignore the first commit. The second one is, in my opinion, a better solution, since it can resolve the commands output readability problem without requiring an additional step.

image

lee-dohm:
Thanks for the contribution 😀

Issue #465, opened 2018-05-02
Contradictory info on default keybindings for Movement and Selection Commands


Botmasher:
According to [Additional Movement and Selection Commands](https://flight-manual.atom.io/using-atom/sections/moving-in-atom/#additional-movement-and-selection-commands), the command `editor:move-to-beginning-of-screen-line` has no keybinding by default.

> the command `editor:move-to-beginning-of-screen-line` is available in the command palette, but it's not bound to any key combination

At the end of that subsection, the same command is listed as having a default keyboard shortcut:

> Here's a list of Movement and Selection Commands that have a keyboard shortcut by default
> ```
> ...
> editor:move-to-beginning-of-screen-line
> ...
> ```

Judging from a fresh install of atom and the atom [keymaps](https://github.com/atom/atom/tree/master/keymaps), it seems like it does not have a default keybinding.

That whole list is mixed in terms of default keybindings. For example, the command `editor:select-to-beginning-of-word` is bound by default, while `editor:select-to-beginning-of-line` is not. Was this meant as a list of movement commands that don't have default keybindings?

I'm happy to make a PR with the update once I understand the intent.

rsese:
Hmmm, yeah - section was added in https://github.com/atom/flight-manual.atom.io/pull/426, I think the list was supposed to be commands with no keybindings yeah @ungb?

ungb:
that is correct. Good catch, I can submit a PR sometime tomorrow, or if someone can get to it before me!

Botmasher:
@ungb Thanks for confirming that - I had the chance to open a pull request [here](https://github.com/atom/flight-manual.atom.io/pull/467).

PR #467, opened 2018-05-09
Clarify movement and selection commands without default keybindings

Botmasher:
I updated "Additional Movement and Selection Commands" to make it clear these commands do not have default keybindings. This comes after checking intentions for the section in issue https://github.com/atom/flight-manual.atom.io/issues/465.

Comparing the Mac, Win32 and Linux lists against current keymaps, I found some key combinations that _are_ bound by default and removed them from the appropriate lists.


ungb:
Thanks for doing this @Botmasher! I just checked too(on windows), I could have sworn at the time I had the list that didn't have key combinations but I can't remember on top of my head.  Either way looks good and thanks again!

Botmasher:
Glad to help!

PR #468, opened 2018-05-09
Update installing-atom.md to include snap

Aaahh:


Arcanemagus:
Thanks for reaching out!

The Atom team does not distribute a Snap image, so I'm not sure where that is coming from, but it is not an official distribution of Atom.

Often these customized versions of Atom are modified versions of the Stable branch of Atom with mismatched versions of built-in components. These updated components are taken from the Beta channel or `master` branch and then injected into the Stable version and a new Atom package is generated. Because of the way Atom is constructed, using these mismatched components can cause mysterious and hard-to-diagnose problems. You can find out more about [why we chose to not support unofficial distributions here](https://github.com/atom/design-decisions/blob/master/unofficial-atom-distributions.md).

PR #474, opened 2018-08-02
Docs for `root-scope-used`

Aerijo:
Initial write up of docs for this activation hook. I'm thinking I should also add more on which hook to pick. Something like
  - `grammar-used` if it is any grammar in a specific package you want to activate on
  - `root-scope-used` if the package providing the grammar does not matter, and you only want to activate when the right language is used. A better explanation on the differences between TextMate and Tree-sitter might be needed, or a link to where the reader can learn more.

Related atom/atom#17521

I also found and fixed a typo, hope that's OK.

### DIFF
--- a/content/hacking-atom/sections/package-word-count.md
+++ b/content/hacking-atom/sections/package-word-count.md
@@ -55,7 +55,8 @@ style sheets your package needs to load. If not specified, style sheets in the `
   *  `core:loaded-shell-environment`,
   *  `language-package-name:grammar-used` (e.g., `language-javascript:grammar-used`), and  
   * `language-scope:root-scope-used` (e.g., `source.js:root-scope-used`).  
-  Using the `root-scope` hook allows your package to be provider agnostic; it will activate when any file using that root scope is opened, regardless of what package is providing the scope. In contrast, the `grammar-used` hook will activate it when any grammar from a particular language package is activated. The `root-scope` hook should be preferred, as it allows users to use different language packages without you needing to add each one to the activation hooks. Language package authors are encouraged to keep scope names consistent with each other to aid in this.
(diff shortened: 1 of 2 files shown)


content/hacking-atom/sections/package-word-count.md
```diff
@@ -51,7 +51,11 @@ style sheets your package needs to load. If not specified, style sheets in the `
...
 * `activationCommands`: an Object identifying commands that trigger your package's activation. The keys are CSS selectors, the values are Arrays of Strings identifying the command. The loading of your package is delayed until one of these events is triggered within the associated scope defined by the CSS selector.  If not specified, the `activate()` method of your main export will be called when your package is loaded.
-* `activationHooks`: an Array of Strings identifying hooks that trigger your package's activation. The loading of your package is delayed until one of these hooks are triggered. Currently, there are two activation hooks: `language-package-name:grammar-used` (e.g., `language-javascript:grammar-used`) and `core:loaded-shell-environment`.
+* `activationHooks`: an Array of Strings identifying hooks that trigger your package's activation. The loading of your package is delayed until one of these hooks are triggered. Currently, there are three activation hooks:
+  *  `language-package-name:grammar-used` (e.g., `language-javascript:grammar-used`),
+  * `language-scope:root-scope-used` (e.g., `javascript:root-scope-used` or `source.gfm:root-scope-used`), and
```
50Wliu: `source.javascript`, no?

Aerijo: That was supposed to be a Tree-sitter example. I chose it because it will eventually be enabled by default. All the more reason for an explanatory paragraph though (or a dedicated section on hooks / the entire package activation process).

Aerijo:
- Tree-sitter has moved back to the original scopes, so just `source.js` and no more `javascript`
- I added an explanation of what each one does.
- The last sentence `Language package authors are encouraged to keep scope names consistent with each other to aid in this.` is not strictly relevant, but I think it helps any budding language package authors (who think it will be fun to write one...) understand why consistency is desirable.

Arcanemagus:
Should the minimum supported version be noted in the documentation in order for package authors to update their `engines` field?

Aerijo:
@Arcanemagus I've never thought about the `engines` field before; I've just used the `">=1.0.0 <2.0.0"` suggestion. I suppose it makes sense though.

PR #478, opened 2018-08-13
Add Tree-sitter grammar section

maxbrunsfeld:
This PR adds a section about adding Tree-sitter grammars to Atom.

[Rendered](https://github.com/atom/flight-manual.atom.io/blob/mb-tree-sitter-grammar-section/content/hacking-atom/sections/creating-a-grammar.md)

/cc @queerviolet @lee-dohm @50Wliu @Aerijo - For some reason, I'm not able to formally 'request a review' from you; but if you have opinions on these docs, could you 👀 this page?

### DIFF
--- a/content/hacking-atom/sections/creating-a-grammar.md
+++ b/content/hacking-atom/sections/creating-a-grammar.md
@@ -1,7 +1,7 @@
 ---
 title: Creating a Grammar
 ---
-### Creating a Tree-sitter Grammar
+### Creating a Grammar
 
 Atom's syntax highlighting and code folding system is powered by [Tree-sitter](http://tree-sitter.github.io/tree-sitter). Tree-sitter parsers create and maintain full [_syntax trees_](https://en.wikipedia.org/wiki/Abstract_syntax_tree) representing your code.
 
--- a/content/hacking-atom/sections/creating-a-grammar.md
+++ b/content/hacking-atom/sections/creating-a-grammar.md
@@ -1,14 +1,20 @@
 ---
-title: Creating a Tree-sitter Grammar
+title: Creating a Grammar
 ---
 ### Creating a Tree-sitter Grammar
 
-[Tree-sitter](http://tree-sitter.github.io/tree-sitter) is a new parsing system that Atom can use for syntax highlighting and code folding. Unlike a TextMate grammar, a Tree-sitter parser creates and maintains a full [_syntax tree_](https://en.wikipedia.org/wiki/Abstract_syntax_tree) representing your code. This syntax tree allows Atom to understand your code more accurately, which has several benefits:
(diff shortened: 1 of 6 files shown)


maxbrunsfeld:
Yeah, if you clone the repo, check out this branch, and run `script/bootstrap` and `script/server`, it'll serve the generated site on port 4000.

queerviolet:
Okay, per discussion with @maxbrunsfeld, I changed the flow and some language to make it clear that Tree-sitter grammars are the preferred way to create grammars, and that TextMate grammars are for legacy use, and we intend to eventually phase them out.

content/hacking-atom/sections/creating-a-grammar.md
```diff
@@ -3,203 +3,292 @@ title: Creating a Grammar
...
+```
 
-**Tip:** Grammar packages should start with _language-_.
+then run the command `npm publish`.
 
-{{/tip}}
+#### The Package
 
-The default package template creates a lot of folders that aren't needed for grammar packages. Go ahead and delete the `keymaps`, `lib`, `menus`, and `styles` folders. Furthermore, in `package.json`, remove the `activationCommands` section. Now create a new folder called `grammars`, and inside that a file called `flight-manual.cson`. This is the main file that we will be working with - start by populating it with a [boilerplate template](https://gist.github.com/DamnedScholar/622926bcd222eb1ddc483d12103fd315). Now let's go over what each key means.
+Once you have a Tree-sitter parser that is available on npm, you can use it in your Atom package. Packages with grammars are usually named starting with _language_. You'll need a folder with a `package.json`, a `grammars` subdirectory, and a single `json` or `cson` file in the `grammars` directory, which can be named anything.
```
lee-dohm: Just to make it more clear, we should probably strengthen the wording:

> Packages with grammars are usually named starting with _language_.

to:

> Packages with grammars are, by convention, always named starting with _language_.

Even if things won't go wrong right now if they don't follow the convention, there are some things within Atom that started as convention but then behavior within Atom started to depend on that convention later, causing confusion.

content/hacking-atom/sections/creating-a-grammar.md
```diff
@@ -3,203 +3,292 @@ title: Creating a Grammar
...
-To start, let's add a basic pattern to tokenize the words `Flight Manual` whenever they show up. Your regex should look like `\bFlight Manual\b`. Here's what your `patterns` block should look like:
 ```coffee
-'patterns': [
-  {
-    'match': '\\bFlight Manual\\b'
-    'name': 'entity.other.flight-manual'
-  }
-]
+id: 'mylanguage'
+name: 'MyLanguage'
```
lee-dohm: Since this is human-readable, should it include spaces? If it cannot or should not include spaces, we should probably specify that it should not.

50Wliu:
It took me a few tries to understand the four different types of values for selectors (that summary right before "specificity").  I think it would help to provide a CSON example of all four types and comment it.

maxbrunsfeld:
Alright, thanks for the feedback everybody. I think I've incorporated it all.

PR #490, opened 2018-10-14
Explain how to open a file at a certain line (and column)

fedelibre:
This PR completes the work started [here](https://github.com/atom/atom/pull/17569).

@lee-dohm please review


Ben3eeE:
@fedelibre Thanks for contributing :bowing_man:

This looks great to me. One thing I would like to see is an example for opening at a specific column as well. This currently mentions that it's optionally allowed but not how it's done.

fedelibre:
@Ben3eeE Ok, done.
The `git grep --column` option has been recently added (git version 2.19)...

Ben3eeE:
Thanks for contributing 🙇 💯

PR #493, opened 2018-10-31
fix calling bootstrap on windows

UziTech:
calling `script/bootstrap` in command line on Windows displays an error:

```
'script' is not recognized as an internal or external command,
operable program or batch file.
```

The solution is to use backslash instead of forward slash `script\bootstrap`

### DIFF
--- a/content/hacking-atom/sections/hacking-on-atom-core.md
+++ b/content/hacking-atom/sections/hacking-on-atom-core.md
@@ -19,18 +19,32 @@ $ git clone git@github.com:<em>your-username</em>/atom.git
 
 From there, you can navigate into the directory where you've cloned the Atom source code and run the bootstrap script to install all the required dependencies:
 
-<span class="platform-mac platform-linux">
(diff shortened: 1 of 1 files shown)


50Wliu:
You should split this up into a Windows section and a Mac/Linux section.

Ben3eeE:
Building Atom seems to have the same issue where it instructs to use `script/build` on all platforms.

UziTech:
Maybe this should change when switching between platforms. I'm not sure how to do that on this repo.

flightmanual

UziTech:
I fixed `script/build` as well. Thanks for the suggestion @Ben3eeE

UziTech:
@Ben3eeE is this still being reviewed?

UziTech:
@Ben3eeE I rebased so this should be able to be merged

Ben3eeE:
Thanks for contributing ✨ Apologies for the delay in merging 🙇

PR #495, opened 2018-10-31
Update installing-atom.md

alevinetx:
The file locations for Windows was incorrect.  It now matches Mac/Linux in that the .atom folder is in the same folder as the app binary.

### DIFF
--- a/content/getting-started/sections/installing-atom.md
+++ b/content/getting-started/sections/installing-atom.md
@@ -168,11 +168,11 @@ To setup Atom in portable mode download the [zip/tar.gz package for your system]
 
 {{#windows}}
 
-Then create a `.atom` directory alongside the directory that contains atom.exe, for example:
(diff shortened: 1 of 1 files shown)


alevinetx:
Okay, here's what I found, as I just checked:

running 1.32.1 on Windows
Paths:
A: \dev\atom
B: \dev\atom\1.32.0
C: \dev\atom\1.32.0\app-1.32.1

atom.exe exists in B.  It turns out, it also exists in C; I didn't see it the other day when I did a quick glance, only saw DLLs.

b: atom.exe = 422k
c: atom.exe = 65M

a\.atom:
running B: fails

b\.atom:
running B succeeds
running C succeeds.

so, it seems we're both correct? :)   The binary in B appears to be a shallow shell that invokes the proper binary in C.  I honestly don't recall how it was originally installed, but I would guess via installer.

I'm fine with closing the PR.  If there's not any documentation noting the 2 different binaries, maybe that could be added somewhere.

PR #497, opened 2018-11-10
Use correct scope name for addInjectionPoint

314eter:
The javascript scope name changed from `javascript` to `source.js` in atom/language-javascript#591. The same change was done in atom/language-javascript@ea37ba47038b9e6f253a8a4b22cf7b611e79bf98.


50Wliu:
Thank you!

PR #501, opened 2018-11-27
pin version of gulp-connect

annthurium:
This remediates the 
vulnerability with `event-stream` outlined in https://github.com/dominictarr/event-stream/issues/116#issuecomment-440927400

### Background

`event-stream` is a dependency of `gulp-connect`.  By [pinning](https://renovatebot.com/docs/dependency-pinning/) the version of `gulp-connect`, now we are no longer including any version of `flatmap-stream` with this code.

I've seen some reports that only `flatmap-stream@0.1.1` contains the maliciously injected code, as part of the minified bundle.  `flight-manual.atom.io` was only ever running `flatmap-stream@0.1.0`, however it's better to be safe than sorry.

Downsides of pinning the version is that we won't get automatic upgrades when new versions of `gulp-connect` are released, and we'll have to manually upgrade. 

### Verification plan

Run local version of server to verify that the site still builds and loads as expected.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,21 +12,21 @@
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.5.tgz",
       "integrity": "sha1-63d99gEXI6OxTopywIBcjoZ0a9I=",
       "requires": {
-        "mime-types": "~2.1.18",
+        "mime-types": "2.1.21",
         "negotiator": "0.6.1"
       },
       "dependencies": {
         "mime-db": {
-          "version": "1.36.0",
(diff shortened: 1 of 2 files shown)


jasonrudolph:
CI is [failing due to broken links](https://travis-ci.org/atom/flight-manual.atom.io/builds/460506677#L838), but those broken links are the result of the missing documentation described in https://github.com/atom/atom/issues/18308, not a bug in the flight manual.

That missing documentation will be resolved by the next Atom release, at which point the links will no longer be broken, and CI should once again be green. 😅

PR #503, opened 2018-12-01
use wget instead of curl for installing Atom

eyelash:
since curl is not installed by default on Ubuntu


lee-dohm:
Is `wget` installed by default on Debian?

PR #504, opened 2018-12-07
Update navigation keybindings for the GitHub package

smashwilson:
To prepare for editable diffs, we've changed some of the navigation keybindings in the GitHub package to use `Cmd-` or `Ctrl-` prefixes. Both the old ones and the new ones should work until editability ships, but it would be good to prepare readers for the change sooner rather than later.

I've updated the appropriate `<kbd>` snippets in the GitHub package docs to reflect this.


Emciu (commented):
Gf

PR #506, opened 2018-12-14
add Cmd+D shortcut to select word

Jetinho:


rsese:
Thanks for the suggestion @Jetinho!  <kbd>cmd+d</kbd> does behave like <kbd>ctrl+shift+w</kbd> (`Editor: Select Word`) initially but it's slightly different since the behavior isn't the same when used multiple times.  <kbd>ctrl+shift+w</kbd> selects the current word only but <kbd>cmd+d</kbd> (`Find and Replace: Select Next`) will select the next occurrence of the currently selected word when used repeatedly.

Additionally, <kbd>cmd+d</kbd> is referenced in a different section where we feel it fits best:

https://flight-manual.atom.io/using-atom/sections/editing-and-deleting-text/#multiple-cursors-and-selections

Because of the above we're going to keep things as is but thanks so much for your contribution!

PR #511, opened 2019-01-19
Change links for forks

SConaway:
Change `$ git clone git@github.com:<em>your-username</em>` to `$ git clone "git@github.com:$(git config --get github.user)"`.

This makes for easier copying and pasting, as one doesn't have to type in their user.

Yes, they need `github.user` set, but isn't this common? I'll add a section on setting it.

### DIFF
--- a/content/hacking-atom/sections/contributing-to-official-atom-packages.md
+++ b/content/hacking-atom/sections/contributing-to-official-atom-packages.md
@@ -1,25 +1,32 @@
 ---
 title: Contributing to Official Atom Packages
 ---
-### Contributing to Official Atom Packages
+
+# Contributing to Official Atom Packages
 
 If you think you know which package is causing the issue you are reporting, feel free to open up the issue in that specific repository instead. When in doubt just open the issue on the [atom/atom](https://github.com/atom/atom) repository but be aware that it may get closed and reopened in the proper package's repository.
 
-#### Hacking on Packages
+## Hacking on Packages
 
-##### Cloning
+### Cloning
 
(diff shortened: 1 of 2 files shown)


rsese:
Thanks @SConaway!  Can you share a small description of the change?

It looks like this could make things a tiny bit easier for people to follow along and they can just copy/paste rather than having to manually enter their username?  But someone would have to have set `github.user` in their Git config yes, otherwise this won't work?

SConaway:
See edit.

rsese:
It looks like a number of other formatting changes were added unrelated to the fork URL change?  It's much easier for us to review pull requests when the change is specific and as described in the title and pull request body (small formatting changes can be fine but it looks like all the headings levels are changed as well as the character used for list items).  As far those changes, that's not something we would accept right now but if we revisit some of those formatting choices in the future, it's something we'll keep in mind.

As far as the fork URL change:

> Yes, they need github.user set, but isn't this common? I'll add a section on setting it.

Hmm, it's not something documented at https://help.github.com/ and I had never heard of that setting as a convention before.  It also isn't a setting that any GitHub app sets as far as I know either (e.g. Desktop and Atom don't set it).  I asked a few other maintainers and they don't have that setting in their Git config either.  

In any case, the github.com documentation documents it similarly where there's just a placeholder for your username (https://help.github.com/articles/configuring-a-remote-for-a-fork/) and with these changes, if someone doesn't have the setting already, they would have to manually type out their username anyways to run the config command.  

If we hear about a lot of confusion about the fork URL we might revisit this, but for now we're going to leave things as they are.  But thanks again for the contribution, we do appreciate it :bow:

PR #519, opened 2019-04-03
Migrate FAQs from https://discuss.atom.io/c/faq to the Flight Manual

lee-dohm:
Still not sure if we should migrate all FAQs or just the more recent ones.


lee-dohm:
@rsese Would you mind taking a look at this to ensure that:

1. I got [all of the FAQs](https://discuss.atom.io/c/faq) that are still applicable and useful
1. There aren't any that you think are not useful anymore

rsese:
:+1:  Did you need me to take a look by a particular day/time?

> Would you mind taking a look at this to ensure that:
>
> * I got [all of the FAQs](https://discuss.atom.io/c/faq) that are still applicable and useful
> * There aren't any that you think are not useful anymore

We have that FAQ audit planned for later in the quarter which would overlap with this :point_up:  Do you think it's fine to just take everything and then I'll audit them later?  Or if you'd rather we do some amount of auditing now for this migration, would you be ok with me doing a high level look through each item and calling out just the most super obvious ones that are no longer useful?

lee-dohm:
That's a good point. I can just merge what I've got now and bring over any other ones that we find we need over time. Then we can do the full audit later.

rsese:
Sounds good :v:

PR #531, opened 2019-05-24
Corrected deactivation advice

### LINKED ISSUE #508: Deactivation advice is incorrect
The current advice on deactivation in the [word-count](https://flight-manual.atom.io/hacking-atom/sections/package-word-count/) page reads as follows:
> This optional method is called when the window is shutting down. If your package is watching any files or holding external resources in any other way, release them here. If you're just subscribing to things on window, you don't need to worry because that's getting torn down anyway.`

This advice is **incorrect**, as deactivation also occurs when the package is disabled, which can happen independently of window shutdown. You should in fact dispose of all subscriptions you're holding on to.

ricfri:
Deactivation advice mentioned in #508 corrected.

Fixes #508.


Arcanemagus:
Thanks!

PR #532, opened 2019-05-28
FAQ updates :pencil:

rsese:
Removed:

* I want Ctrl-Tab to work like it used to. How do I do that?
* Where did the “Show Tree View On Right Side” option go?

These items have been less relevant with each release - how things used to work relative to 1.14 and 1.17.

Added:

* How do I accept input from my program or script when using the script package?
* I have a question about a specific Atom community package. Where is the best place to ask it?

First question still gets asked about and "python input" is a searched for a lot in the message board.  Second question is something we reference/will link to when we find someone is having an issue with a community package.


rsese:
Also added:

* What platforms does Atom run on?
* How can I contribute to Atom?
* Why does Atom collect usage data?

Pulled these in from https://atom.io/faq - we'll redirect this page to the Flight Manual FAQ

PR #534, opened 2019-06-03
Document scope.name:root-scope-used activation hook

Arcanemagus:
### Requirements for Contributing Documentation

* Fill out the template below. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* The pull request must only contribute documentation (for example, markdown files or API docs). To contribute other changes, you must use a different template. You can see all templates at https://github.com/atom/atom/tree/master/.github/PULL_REQUEST_TEMPLATE.

### Description of the Change

Added to Atom since v1.30.0 in https://github.com/atom/atom/pull/17521, the `scope.name:root-scope-used` activation hook wasn't currently documented anywhere other than the PR that added it.

This adds it as an example and gives some explanation for the other two hooks as well.

### Release Notes

N/A

----

Closes #474.


Arcanemagus:
Hmmm, this is actually an alternative to #474, will try to get a consensus on which should be merged.

PR #539, opened 2019-06-26
⬆️ Upgrade to Node 8

jasonrudolph:
Upgrade to Node 8 so that we can use async/await. :pray:


jasonrudolph:
@lee-dohm: Would you mind reviewing this pull request when you have a moment?

PR #541, opened 2019-06-27
Add Atom's API docs to the Flight Manual

jasonrudolph:
This pull request aims to bring the Atom API reference docs into the Flight Manual. These docs currently reside at https://atom.io/docs/api, and we'd like for them to move them into this repository so that all of Atom's docs are consolidated in a single place.

### TODO

- [x] Enhance the build to generate API docs at `/api/:version/:class` (for example: `/api/v1.38.2/AtomEnvironment`)
- [x] Enhance the search index to include the API docs
    - [x] instance methods (aeee307bf27aaf0a2e611084e86cb8e5d99c0202)
    - [x] instance properties (fb77ad9f30c6780980f6a77cf6fad982508c9b5a)
    - [x] class methods (4599da9b4c3b1401858eb6a66bca4ae17e143592)
    - [x] the class itself (0a417a8d00349731074246a4de21240f89b382eb)
- [x] Show the list of classes in the left sidebar (to match current behavior on https://atom.io/docs/api)
- [x] When navigating to the root URL (`/api`), show the docs for the current stable version (https://github.com/atom/flight-manual.atom.io/compare/387eeee4488f13be1eb610370ef495d9bbce98b9...de1e02a850721dfcf1cf2a5e8ee322c09b92b5eb)
- [x] Provide drop-down menu for choosing the Atom version whose docs you want to view
- [x] Compare the output side-by-side with the content at https://atom.io/docs/api and fix any bugs
    - [x] "Essential" classes should include an "Essential" label at the top of the page [[screenshot](https://user-images.githubusercontent.com/2988/60625837-f0bc6d00-9db6-11e9-83f4-0e78c12d96cb.png)] (60234aa2018795ae0b3b0d03a87f12630c01d051)
    - [x] Code samples should be rendered with syntax highlighting [[screenshot](https://user-images.githubusercontent.com/2988/60625838-f0bc6d00-9db6-11e9-8555-5f258c24dbe2.png)]
    - [x] Optional parameter should have an "optional" badge [[screenshot](https://user-images.githubusercontent.com/2988/60625840-f1550380-9db6-11e9-821c-d75d31511e15.png)] (5a0f3900776dec193b301024f42502569949cd61)
    - [x] "Extended" methods should be grouped together in an "Extended Methods" subsection, and they should be hidden by default [[screenshot](https://user-images.githubusercontent.com/2988/60626493-ef8c3f80-9db8-11e9-9230-5707698a8d02.png)]

/cc @lee-dohm @as-cii


jasonrudolph:
> * [ ]  When navigating to the root URL (`/api`), show the docs for the current stable version

@lee-dohm @as-cii: I've just about got this ☝️ working locally. I've gotta head out for dinner, but I'm hoping to get this pushed up tomorrow morning.

lee-dohm:
> * [ ] Show the list of classes in the left sidebar (to match current behavior on https://atom.io/docs/api)

I got this working for both the main table of contents at `/` and in the sidebar of each page. I sandwiched the API docs between the chapters and the appendices. I fiddled around with the memoization to speed things up and it's pretty quick for going through 140+ versions of API docs.

Right now though, the class links all refer to `/api/latest/ClassName/index.html` so either we'll need to fiddle with the `routes` rules in `Rules` or, if that won't work, I can quickly redo it to put the latest version number in the URL.

lee-dohm:
Updated the links to use a hardcoded version.

jasonrudolph:
> - [ ] Compare the output side-by-side with the content at https://atom.io/docs/api and fix any bugs

I've compared the current content at https://atom.io/docs/api to the content rendered by this branch, and I've updated the issue body to list the differences that I've noticed so far.

jasonrudolph:
@lee-dohm: I think this is looking pretty good now. Would you mind trying out these changes to see if there's anything that's essential to change before we ship?

Note: Given the size of this diff and how long this branch has been in progress, I think we should lean heavily toward getting to a reasonable production "v1" of these changes as soon as possible and deferring any enhancements to separate pull requests. 😇

lee-dohm:
I'll take a look at this first thing in my morning tomorrow. Thanks so much for all the work driving this forward @jasonrudolph and @as-cii 🙇 I agree that a reasonable v1 is the appropriate action here.

lee-dohm:
This looks great! I've marked the PR as ready for review while I dig deeper into it.

calebmeyer:
From taking a quick look at the website:
- I like the search, although I wish the indexing were breadth first instead of depth first. Searching "marker" gets you lots of methods under MarkerLayer, but nothing about DisplayMarker.
- I think our sidebar nav is getting a bit long. I'd almost rather see the headers be collapsible, even if they start expanded. One of the things I liked about the separate API docs was that I could see quite a few of the classes without any scrolling. Now even veterans have to scroll quite a bit to get past the sections on installing atom, and editing text.
- I didn't see any straight up bugs from clicking around. The mac/windows/linux selector works, all the highlighting looks right, all the links worked, including source code links. :+1:

layouts/search.json
```diff
@@ -0,0 +1,58 @@
+{
```
calebmeyer: Why don't we append `.erb` on these files? It would make them more likely to syntax highlight correctly in whatever editor you're using, including github web.

lib/temp.rb
```diff
@@ -0,0 +1,34 @@
...
+    value
+  end
+end
+
+def temp_read(name)
+  lines =
+    File.readlines(temp_join(name))
+    .map { |line| line.strip }
+
+  lines.length > 1 ? lines : lines.first
```
calebmeyer: I'd call it generally an anti-pattern to return two different types from a function. Here we're returning either an array of strings or a string. Did you mean to join the lines with "\n"?
```ruby
def temp_read(name)
  File.readlines(temp_join(name)).map(&:strip).join("\n")
end
```

calebmeyer: [Reference](https://www.brianstorti.com/understanding-ruby-idiom-map-with-symbol/) for the `&:` shorthand, in case anyone hasn't seen it before.

lib/temp.rb
```diff
@@ -0,0 +1,34 @@
+
+def temp_exists?(name)
+  File.exists?(temp_join(name))
+end
+
+def temp_join(name)
```
calebmeyer: You may be able to avoid a few of these methods with Ruby's [Tempfile](http://ruby-doc.org/stdlib-2.6.3/libdoc/tempfile/rdoc/Tempfile.html) class.

layouts/api-class.html
```diff
@@ -0,0 +1,38 @@
...
+
+        <div class="toc">
+          <%= renderp '/includes/toc.html', :heading => 'h4', :title => @item[:title], :version => version %>
+        </div>
+
+        <div class="content-column">
+          <% section = lookup_section('toc', @item[:title]) %>
+          <h1 class="document-title">
+            <%= renderp '/includes/platform-nav.html' %>
+            <!-- <a class="improve-link"
```
calebmeyer: commented "code"

assets/javascripts/search.js
```diff
@@ -0,0 +1,218 @@
...
+      debounce(() => {
+        this.quickSearchResults.hide();
+      })
+    );
+  }
+
+  bindQuicksearchMousedown() {
+    $(".autocomplete-results a").each(function(_, el) {
+      $(el).bind("mousedown", function(event) {
+        event.preventDefault();
```
calebmeyer: Just preventing the default and returning? I'm assuming this is so the anchors don't try to visit their hrefs. Should we use a different element like `<button>` in that case to be more semantically accurate?

assets/javascripts/search.js
```diff
@@ -0,0 +1,218 @@
+$(document).ready(function() {
```
calebmeyer: There's a shorthand for this. Any function given directly to the jQuery function is called on document ready.

[Reference](https://api.jquery.com/jQuery/#jQuery3)

lib/temp.rb
lee-dohm: There's a scenario where we need to memoize an array of strings and memoize an individual string, at least when I wrote the first version. I'm not sure if that's necessary anymore.

layouts/search.json
lee-dohm: Fair point. Would you mind opening an issue for this?

assets/javascripts/search.js
lee-dohm: I think that's a valid change but probably should be put off until after this is deployed. Would you mind opening an issue?

layouts/search.json
calebmeyer: Opened #549

lib/temp.rb
lee-dohm: Inlined all the functions that aren't used outside of `temp.rb` and `temp_memoize` now always returns an array.

lib/temp.rb
lee-dohm: Decided to stick with `File` for now because I'm not sure if the memoization would work right with `Tempfile`s out-of-scope semantics.

assets/javascripts/search.js
calebmeyer: Opened #550

jasonrudolph:
Thanks for the review! I'm gonna roll this out now.

Despite all our testing and code review, I anticipate us discovering at least _some_ issues in real-world use due to the sheer magnitude of this change. To minimize the impact, we'll keep an eye on issues filed over the next few days, and we'll try to respond as quickly as possible to any problems that come up.

PR #512, opened 2019-02-05
Convert CoffeeScript code examples to JavaScript

lee-dohm:
This intentionally leaves alone:

* Examples having to do with configuration files
* The chapter on upgrading your package to the Atom v1.0 API (since anything pre-1.0 was in CoffeeScript anyway)

PR #559, opened 2019-08-31
Fix `this` context in JS snippets & increase consistency

Aerijo:
RIP template links.

Fixes https://github.com/atom/atom/issues/19855

## Description

Changes the function kind so that the context of `this` is now `TextEditorElement` and `this.getModel()` works. Also changed the following snippet to use `this.getModel()` as well.

Semicolons introduced for consistency with Atom code base (and the other snippets in here I came across).

## Alternatives

`TextEditorElement` is [undocumented](https://github.com/atom/atom/pull/13753). Technically could replace `this.getModel()` with `atom.workspace.getActiveTextEditor()`, to avoid an undocumented method call. However, getModel feels more correct in terms of linking the event to the right editor; it is possible for `atom.workspace.getActiveTextEditor()` to return a different editor, or none at all, whereas we can (probably) safely assume `this.getModel()` will always return the correct editor.

### DIFF
--- a/content/behind-atom/sections/keymaps-in-depth.md
+++ b/content/behind-atom/sections/keymaps-in-depth.md
@@ -80,11 +80,11 @@ Commands are custom DOM events that are triggered when a key combination or sequ
 
 ```javascript
 atom.commands.add('atom-text-editor', {
-  'user:insert-date': (event) => {
-    const editor = this.getModel()
-    return editor.insertText(new Date().toLocaleString())
+  'user:insert-date': function (event) {
+    const editor = this.getModel();
+    return editor.insertText(new Date().toLocaleString());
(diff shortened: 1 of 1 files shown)


rsese:
Thanks @Aerijo :+1:  @lee-dohm - I remember you converted examples to JavaScript in https://github.com/atom/flight-manual.atom.io/pull/512, does this look good to you?

content/behind-atom/sections/keymaps-in-depth.md
```diff
@@ -80,11 +80,11 @@ Commands are custom DOM events that are triggered when a key combination or sequ
```
molodou0: Дубликат из #

PR #561, opened 2019-09-08
Update find-and-replace.md

selfpublish:
### Description of the Change

Previously, it was unclear to me how to specify a single folder to search when you had multiple folders added to a project. It was not obvious that you could actually type the name of a folder in that box at the bottom (under 'Replace in Project') to refine your search. The screenshot should help others see this option more clearly.

### Release Notes

Added a screenshot to clarify how to perform a 'find and replace' within a single folder.

### DIFF
--- a/content/using-atom/sections/find-and-replace.md
+++ b/content/using-atom/sections/find-and-replace.md
@@ -34,6 +34,12 @@ You can limit a search to a subset of the files in your project by entering a [g
 
 When you have multiple project folders open, this feature can also be used to search in only one of those folders. For example, if you had the folders `/path1/folder1` and `/path2/folder2` open, you could enter a pattern starting with `folder1` to search only in the first folder.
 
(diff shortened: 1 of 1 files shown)


selfpublish:
Thanks.

It wasn't clear to me that you actually could type any text in the bottom
box (to search within a directory). I kept trying to modify my search term
following the convention referenced as a text prompt "File/directory
pattern" - might be simply because that last line is narrower than the
first two. I only discovered how to do it accidentally. Maybe it's more of
a UX issue.

[image: image.png]

My best,
Brian

Arcanemagus:
Since the paragraph I mention explicitly says to put the directory `into the "File/Directory pattern" text box`, I'm going to go ahead and close this 😉.

If you come up with wording there that might be clearer we would be glad to review another PR though!

Thanks for taking the time to file this PR.

PR #564, opened 2019-09-21
Naming conventions for syntax scopes

chbk:
## Motivations for this PR

The lack of documentation on syntax scopes has been an [issue](https://github.com/atom/atom/issues/8430) for many years and still hinders the progress of contributors who wish to create themes and grammar packages for Atom. An [attempt](https://github.com/atom/flight-manual.atom.io/pull/114) at fixing this has been made in 2015 but hasn't evolved since.

To resolve this ongoing issue, I propose this documentation based on [Textmate scopes](https://macromates.com/manual/en/language_grammars#naming_conventions). Most of the conventional scopes have been preserved for backwards-compatibility with existing themes. However, I have also introduced new scopes and removed redundant ones. This documentation aims to be:
- Explicit. The more language-agnostic scopes we provide, the better we prevent scope conflicts between language grammars and the easier it becomes to create universal themes.
- Consistent. Having the same elements scoped the same way throughout a document promotes coherent coloring and makes code intelligible.

This is a work in progress. Any suggestions to improve the naming conventions are welcome.

#### `symbol`
Symbols are non-alphabetic tokens used as reserved words. Most syntax themes highlight symbols the same way regardless of their more specific scopes. This scope makes it easier for such themes to target all symbols globally.

#### `storage.definition`
The `storage.type` scope is too general. The lack of specificity results in this scope being misused, e.g. `let` scoped as `storage.type` in JavaScript, or just blatantly ignored, e.g. `func` scoped as `keyword.function` in Go or `enum` scoped as `keyword.control` in C. An additional more specific scope to differentiate actual variable types from definition keywords is needed.

#### `storage.type.fundamental`
We should provide a way to differentiate built-in fundamental types from classes used as types. Generally, built-in types can be divided into two categories: primitive types, which are fundamental types that cannot be decomposed into smaller components, e.g. `int` and `boolean`, and composite types, which are fundamental types constructed from primitive types, e.g. `map` and `dict`. However, classes can also be considered composite types, so adding a `composite` scope will not help distinguish built-in types from other classes. Adding a `fundamental` scope helps target built-in fundamental types meaningfully.

#### `punctuation`
Punctuation is part of language grammars and should not be neglected.

#### __More specific scopes__
Every language has their own vocabulary, and that creates conflicts between scopes. Let's consider the `module` scope as an example. Depending on the language, a module can be a namespace (Rust), an imported package (Python), an abstract class (Ruby)... The `module` scope becomes meaningless for syntax themes without the language context. We can resolve this by providing pre-defined scopes that call similar elements by the same name, regardless of the language. In this case, these scopes would be `namespace` for a collection of entities, `package` for an imported element, and `class.abstract` for an abstract class.

#### `entity`
The original definition from the Textmate docs reads "An entity refers to a larger part of the document [...]. We do not scope the entire entity as entity.* (we use meta.* for that). But we do use entity.* for the placeholders in the larger entity [...]". This leads to `entity` being solely used to scope identifiers and re-used whenever the identifiers re-appear in the document. Hence, the practical definition for `entity` is simply "a named element".

#### `support`
Using this scope should not exempt one from also using other relevant scopes. Therefore `support` is no longer its own standalone category but is appendable to any other scopes when appropriate. Furthermore, the original definition for this scope was "things provided by a framework or library". But in practice it is mostly used for built-in entities rather than imported ones. This justifies changing the definition to "a built-in or imported or conventional entity which can usually be redefined".

#### `invalid`
An invalid token can either be a deprecated entity or an illegal token. Similarly to `support`, it should be used with other relevant scopes when applicable. Often these scopes would be `entity` or `keyword` or `punctuation`.

#### `variable`
A variable is an entity. Similarly to `support` it should not be used as a standalone scope.

#### `other`
This scope is used too liberally and contributes no information.

#### `name` in `entity.name`
Adding `name` to `entity` is redundant as the latter is solely used to scope named elements.

#### `java`, `python`, etc.
Trailing scopes identifying the language are redundant as this information is provided by the root scope, e.g. `source.java`.

## Amendments to the Documentation

I propose adding the following section to the [Hacking Atom](https://flight-manual.atom.io/hacking-atom/) chapter of the Atom flight manual. I use "class" instead of "scope" below to conform to the terminology used in existing documentation.

### Naming Conventions

Naming conventions ensure that every language is correctly highlighted by different themes. When creating a language grammar, use these conventions to decide which syntax classes correspond to your syntax-tree nodes. When creating a syntax theme, use these conventions to decide which syntax classes correspond to your theme colors.

We detail the naming conventions below, organized into nested categories. In a language grammar, multiple classes can be applied to a syntax-tree node. A nested class is appended to its parent class by adding a `.` between them (as in `keyword.operator`). Shared classes indicated in brackets (like `[symbol]`) can be appended subsequently. The order in which classes are appended does not affect the resulting highlighting. However, to preserve readability, nested classes should be appended following their nesting order (as in `keyword.operator.comparison`) and shared classes should be added last (as in `keyword.operator.comparison.symbol`). General classes which are sufficient enough for coherent highlighting are indicated in bold. Other more specific classes fine-tune the highlighting. This list is extensive but not exhaustive. You may create additional specific classes (like`less-than` in `keyword.operator.comparison.less-than.symbol`) if they clarify your grammar and do not introduce conflicts with existing classes.

- __`keyword`__ — A reserved word that may not be used as an identifier.
    - __`symbol`__ — A non-alphabetic token used as a reserved word.
    - __`control`__ — A keyword used to structure a program.
        - `condition` — A condition keyword.
          Examples: `if`, `else`, `elif`.
        - `loop` — A loop keyword.
          Examples: `for`, `for`...`in`, `for`...`of`, `do`, `while`.
        - `exception` — An exception keyword.
          Examples: `try`, `catch`, `finally`.
        - `result` — A result keyword.
          Examples: `throw`, `raise`, `break`, `continue`, `pass`, `return`, `yield`.
        - `package` — A keyword for imports or exports.
          Examples: `import`, `from`, `include`, `export`, `require`.
        - `directive` `[package]` — An instruction given to the compiler.
          Examples: `#include`, `#define`, `#ifdef`, `using`, `package`, `use strict`.
        - `evaluate` — A keyword used to evaluate an expression.
          Examples: `assert`, `with`...`as`.
        - `anchor` __`[symbol]`__ — An anchor (RegExp).
          Examples: `^`, `$`, `\b`, `\B`.
    - __`operator`__ — A keyword used for an operation.
        - `logical` __`[symbol]`__ — A logical operator.
          Examples: `and`, `not`, `or`, `!`, `&&`, `||`.
        - `ternary` __`[symbol]`__ — A ternary condition operator.
          Examples: `?`, `:`.
        - `wildcard` __`[symbol]`__ — A wildcard operator, usually used in imports.
          Example: `*`.
        - `assignment` __`[symbol]`__ — An assignment operator.
          Examples: `=`, `:=`.
        - `compound` `[assignment]` __`[symbol]`__ — A compound assignment operator.
          Examples: `+=`, `-=`, `*=`, `%=`.
        - `comparison` __`[symbol]`__ — A comparison operator.
          Examples: `==`, `<`, `>`, `!=`, `in`.
        - `arithmetic`  __`[symbol]`__ — An arithmetic operator.
          Examples; `+`, `-`, `/`, `*`.
        - `reference` __`[symbol]`__ — A reference operator.
          Examples: `&`.
        - `dereference` __`[symbol]`__ — A dereference operator.
          Example: `*`.
        - `bitwise` __`[symbol]`__ — A bitwise operator.
          Examples: `<<`, `>>`, `|`, `&`, `^`, `~`.
        - `increment` __`[symbol]`__ — An increment operator.
          Example: `++`.
        - `decrement` __`[symbol]`__ — A decrement operator.
          Example: `--`.
        - `delete` — A deletion operator.
          Examples: `del`, `delete`.
        - `instance` — An instantiation operator.
          Example: `new`.
        - `probe` — An operator to inquire about the properties of an element.
          Examples: `sizeof`, `typeof`, `instanceof`.
        - `combinator` __`[symbol]`__ — A combinator operator (CSS).
          Examples: `>`, `+`, `~`.
        - `quantifier` __`[symbol]`__ — A quantifier operator (RegExp).
          Examples; `?`, `*`, `+`, `{1,2}`.

- __`storage`__ — A variable type or a storage keyword.
    - __`modifier`__ __`[keyword]`__ __`[symbol]`__ — A reserved word used to detail an implementation.
      Examples: `static`, `final`, `extends`, `register`, `extern`, `throws`, `get`, `global`.
        - `flag` — A flag (RegExp).
          Example: `i` in `(?i)`.
    - __`definition`__ __`[keyword]`__ __`[symbol]`__ — A reserved word used to define or declare an entity.
        - `variable` — A variable definition or declaration.
          Examples: `let`, `var`, `const`.
        - `function` __`[symbol]`__ — A function definition or declaration.
          Examples: `fn`, `func`, `function`, `def`, `lambda`, `=>`.
        - `class` `[abstract]` `[module]` `[interface]` — A class definition or declaration.
          Example: `class`.
        - `enumeration` — An enumeration definition or declaration.
          Example: `enum`.
        - `structure` — A structure definition or declaration.
          Examples: `struct`, `type`...`struct`.
        - `union` — An union definition or declaration.
          Example: `union`.
        - `alias` — An alias definition or declaration.
          Example: `typedef`.
        - `namespace` `[module]` — A namespace definition or declaration.
          Examples: `namespace`,  `mod` (Rust).
    - __`type`__ __`[entity`__ __`[class]`__ `[structure]` `[...]`__`]`__ __`[support]`__ — The type of a variable.
      Examples: `MyClass`, `String`, `List`.
        - __`fundamental`__ — A primitive or composite fundamental type.
          Examples: `void`, `char`, `signed`, `short`, `int`, `bool`, `rune`, `list`, `dict`.
        - `generic` — A placeholder for a type.
          Example: `class Class<generic> {}`.
            - __`language`__ `[wildcard]` __`[symbol]`__ — A reserved word that acts as a type.
              Example: `?` in `List<?> list`.

- __`entity`__ — A named element.
    - `immutable` — An entity whose properties or value cannot be changed.
      Examples: `const immutable`, `final immutable`.
    - `parameter` — A parameter in a definition.
      Examples: `fn function(parameter: type) {}`, `class Class<parameter> {}`.
    - `argument` — An argument in a function call or class instantiation.
      Examples: `instance.method(argument)`, `new Class(argument)`.
    - `definition` — An entity that is being defined or declared.
      Examples: `variable` in `let variable`, `function` in `def function()`.
    - `call` — An entity that is being called.
      Examples: `function` in `function(argument)`, `Class` in `new Class(argument)`.
    - __`support`__ — A built-in or imported or conventional entity which can usually be redefined.
      Examples: `len`, `print`, `String`, `Override`, `self`, `cls`.
    - __`variable`__ __`[support]`__ `[parameter]` `[argument]` `[immutable]` `[definition]` — A variable.
        - __`language`__ __`[symbol]`__ — A reserved word that acts as a variable.
        Examples: `this`, `super`, `arguments`, `@`.
    - __`function`__ __`[support]`__ `[argument]` `[definition]` `[call]` — A function.
        - __`type`__ __`[fundamental]`__ — A function to construct a typed element.
          Examples: `int`, `char`, `bool`, `dict`, `matrix`, `list`.
        - `cast`: A function for type casting.
          Examples: `as.matrix`, `toInt`.
        - `decorator` — A decorator function.
          Examples: `@staticmethod`, `@app.route`.
    - __`class`__ __`[support]`__ `[argument]` `[definition]` `[call]` — A class.
        - `inherited` — An inherited class.
        - `generic` `[parameter]` — A generic class.
          Examples:  `<T>`, `<E>`.
        - `exception` — An exception.
        - `abstract` — An abstract class.
            - `module` — A module (Ruby).
            - `interface` — An interface (Java, Go).
    - `enumeration` `[definition]` — An enumeration.
      Example: `enum enumeration{red, green, blue}`.
    - `structure` `[definition]` — A structure.
      Examples: `struct Structure`, `type Structure struct {}`.
    - `union` `[definition]` — An union.
      Example: `union Union {}`.
    - `alias` `[definition]` — An alias.
      Example: `typedef int Alias`.
    - `annotation` __`[support]`__ `[definition]` — An annotation.
      Examples: `@Override`, `@Deprecated`.
    - `namespace` `[definition]` — A namespace.
      Example: `namespace Namespace {}`.
        - `module` — A module (Rust).
          Example: `module::function()`.
    - `package` — A package.
      Example: `from package import entity`.
    - `key` — A key in a key-value mapping.
      Example: `{key => value}`.
    - `method` `[definition]` `[call]` — A method.
      Examples: `{method: (parameter) => value}`, `object.method()`.
    - `property` `[definition]` — A property.
      Examples: `{property: value}`, `object.property`.
    - __`tag`__ — A tag (HTML).
      Examples: `body`, `div`, `input`.
    - __`attribute`__ — An attribute (HTML).
      Example: `<tag attribute=value>`.
    - __`selector`__ `[tag]` `[id]` `[class]` — A selector (CSS).
      Examples: `#id`, `.class`.
    - __`pseudo-class`__ — A pseudo-class (CSS).
      Examples: `:hover`, `:not`.
    - __`pseudo-element`__ — A pseudo-element (CSS).
      Examples: `::before`, `::after`.

- __`constant`__ `[argument]` — A literal constant other than a string.
    - `character` — A character.
      Example: `&lt`.
        - `escape` — A character escaped with a backslash.
          Example: `\"`.
            - `unicode` — An unicode character.
              Example: `\u2661`.
            - `whitespace` — A whitespace character.
              Examples: `\t`, `\f`.
                - `newline` — A newline character.
                  Examples: `\n`, `\r`.
        - `class` — A character class (RegExp).
          Examples: `.`, `\d`, `\W`, `\s`.
    - __`language`__ __`[symbol]`__ — A reserved word that acts as a constant.
        - `boolean` — A boolean.
          Examples: `true`, `false`.
        - `nothing` — A null value.
          Examples: `None`, `null`, `nil`.
        - `undefined` — An undefined value.
          Example: `undefined`.
        - `ellipsis` __`[symbol]`__ — An ellipsis (Python).
          Example: `...`.
        - `numeric` — A numeric word.
          Example: `Infinity`.
    - __`numeric`__ — A number.
        - `integer` — An integer.
          Example: `2`.
        - `float` — A decimal number.
          Example: `.17`.
        - `hexadecimal` — A hexadecimal number.
          Example: `0x29`.
        - `size` — A size (CSS).
          Examples: `67%`, `97px`, `127pt`, `157em`.
    - `color` — A color (CSS).
      Examples: `crimson`, `#a35`.
    - `font` — A font (CSS).
      Examples: `Helvetica`, `Times New Roman`.
    - `style` — A style (CSS).
      Examples: `break-word`, `solid`, `absolute`.

- __`string`__ `[argument]` — A string.
    - `immutable` — A string whose value cannot be changed.
      Example: `:immutable` (Ruby).
    - `quoted` — A quoted string.
      Examples: `"string"`, `'string'`.
    - `unquoted` — An unquoted string.
      Example: `'key': unquoted`.
    - `heredoc` — A here document.
      Example: `<<EOF A multiline here document EOF`.
    - __`regexp`__ — A regular expression.
      Example: `/^regexp$/`.

- __`punctuation`__ — A punctuation mark.
    - `definition` — Punctuation used to define an element.
        - `function` — Punctuation for a function.
          Examples: `{`, `}`.
        - `decorator` — Punctuation for a decorator.
          Example: `@`.
        - `annotation` — Punctuation for an annotation.
          Example: `@`.
        - `directive` — Punctuation for a directive.
          Example: `#`.
        - `newline` — Punctuation for a newline.
          Example: `\`.
        - `parameters` — Punctuation for parameters.
          Examples: `(`, `)`.
        - `condition` — Punctuation for a condition.
          Examples: `(`, `)`.
        - `expression` — Punctuation for an expression.
          Examples: `(`, `)`.
        - `array` — Punctuation for an array.
          Examples: `[`, `]`.
        - `list` — Punctuation for a list.
          Examples: `[`, `]`.
        - `dictionary` — Punctuation for a dictionary.
          Examples: `{`, `}`.
        - `string` — Punctuation for a string.
          Examples: `"`, `'`.
        - `comment` — Punctuation for a comment.
          Examples: `//`, `#`.
        - `interpolation` — Punctuation for an interpolation.
          Examples: `#{`, `}`.
        - `placeholder` — Punctuation for a placeholder in a string.
          Examples: `{`, `}`.
        - `type` — Punctuation for a type.
          Examples: `<`, `>`.
        - `tag` — Punctuation for a tag (HTML).
          Examples: `<`, `/>`.
        - `pseudo-class` — Punctuation for a pseudo-class (CSS).
          Example: `:`.
        - `pseudo-element` — Punctuation for a pseudo-element (CSS).
          Example: `::`.
        - `selector` — Punctuation for a selector (CSS).
          Examples: `.`, `#`.
        - `group` — Punctuation for a group (RegExp).
            - `capture` — Punctuation for a capturing group.
              Examples: `(`, `)`.
                - `named` — Punctuation for a named group.
                  Examples: `(?P`, `)`.
            - `non-capture` — Punctuation for a non-capturing group.
              Examples: `(?:`, `)`.
            - `look-ahead` — Punctuation for a look-ahead.
              Examples: `(?=`, `)`.
            - `look-behind` — Punctuation for a look-behind.
              Examples: `(?<=`, `)`.
            - `negative-look-ahead` — Punctuation for a negative look-ahead.
              Examples; `(?!`, `)`.
            - `negative-look-behind` — Punctuation for a negative look-behind.
              Examples: `(?<!`, `)`.
        - `set` — Punctuation for a set (RegExp).
          Example: `[`, `]`
        - `negative-set` — Punctuation for a negative set (RegExp).
          Example: `[^`, `]`
        - `flag` — Punctuation for a flag (RegExp).
          Example: `(?`, `)`.
        - `range` — Punctuation for a range (RegExp).
          Example: `-` in `a-z`.
        - `markup` — Punctuation for text styling (Markdown).
          Examples: `_`, `*`, `~`, `#`, `-`. `1.`.
    - `separator` — Punctuation used to separate entities.
      Examples: `,`, `:`.
    - `accessor` — Punctuation used to access contained entities.
      Examples: `.`, `::`, `[`, `]`.
    - `terminator` — Punctuation used to terminate a statement.
      Example: `;`.

- __`markup`__ — Parts of a document which are stylized (Markdown).
    - `underline` — Underlined text.
    - `bold` — Bolded text.
    - `italic` — Italic text.
    - `strike` — Striked-through text.
    - `heading` — A heading.
    - `list` — A list item.
        - `ordered` — A ordered list item.
        - `unordered` — An unordered list item.
    - `quote` — A quote.
    - `raw` — Raw unformatted text or code.
        - `inline` — Inline raw text.
        - `block` — Block raw text.
    - `critic` — A critic.
        - `addition` — An addition.
        - `deletion` — A deletion.
        - `substitution` — A substitution.
        - `highlight` — A highlight.

- __`comment`__ — A comment.
    - `inline` — An inline comment.
      Example: `# comment`.
    - `block` — A multi-line comment.
      Example: `/* ... */`.

- __`invalid`__ __`[entity]`__ __`[keyword]`__ __`[punctuation]`__ — An invalid token.
    - `deprecated` — A deprecated entity or reserved word which should no longer be used.
    - `illegal` — An illegal token which doesn't belong there.

- `meta` — A larger part of the document.
  - `function` — A function definition block.
  - `class` — A class definition block.


chbk:
The naming conventions are fairly complete now. The next step will be to fork existing language grammar packages to implement the conventions and test them with default syntax themes. Comments and contributions are welcome.

The following tree-sitter grammar packages are considered for testing:
- [TODO] [language-c](https://github.com/atom/language-c)
- [TODO] [language-go](https://github.com/atom/language-go)
- [WIP] [language-python](https://github.com/atom/language-python)
- [TODO] [language-ruby](https://github.com/atom/language-ruby)

chbk:
A sneak peek for language-python before (left) and after (right) implementing the conventions.

<p align="center"></p>

__Achievement unlocked:__ Consistency.

adrian5:
Awesome man, doing god's work!

As an aside, I'm wondering if we could add a repository that contains complete code snippets of all (possible) programming languages, covering the full range of available syntax. This would be a boon to syntax-theme developers, even if the syntax-scopes were a moving target.

chbk:
That would be nice, although covering every programming language would be tedious. I'm actually creating snippets to test the grammars listed above and I intend to include them in upcoming pull requests. If you want to create a central repository for code snippets I'd be glad to contribute mine.

Another valuable resource for theme developers is the [syntax theme template](https://github.com/atom/apm/tree/master/templates/theme), which should be updated once the conventions are stable. Ideally, just a couple tweaks to the template should produce the desired coloring consistently across every language. That means a developer would only need to test his theme in a couple languages to be confident it works the same way in every language.

allefeld:
Hi @chbk, thanks for doing this documentation work, it is dearly needed!

What I'm wondering: Atom's syntax scopes are derived/related/inspired (?) by those of TextMate, right? Now, just looking at the top level, [TextMate](https://macromates.com/manual/en/language_grammars#naming_conventions) has 11 "root groups", and you list 10 top-level categories for Atom:

```
             TextMate  Atom
comment          +       +
constant         +       +
entity           +       +
invalid          +       +
keyword          +       +
markup           +
meta             +       +
storage          +       +
string           +       +
support          +
variable         +
punctuation              +  
text                     +
```

Therefore TextMate has 3 "root groups" not in Atom (`markup`, `support`, `variable`) and Atom has 2 top-level categories not in TextMate (`punctuation`, `text`). What is the reason for this difference? If I develop a syntax theme for Atom, should I simply forget about TextMate?

jeff-hykin:
Many of the challenges with existing Textmate scopes are the problems of gray areas.

For example consider the 'variadic sizeof' in C++ which is `sizeof...`. It's an operator, but it is not alphanumeric and it's also not symbolic. It's kind of both.

There's many other similar issues such as lambdas, is a lambda a function definition or a variable? It's both.

Atom can theoretically handle these things quite well because of its use of CSS scopes which don't care about ordering. For example having both a `variable` and a `function` scope/class will allow either of them to be matched, and allow both to be matched with `.variable.function`. In normal Textmate this is nearly an unsolvable problem because of 1. strict ordering, 2. priority given to lowest match, and 3. Lack of a direct-child operator. 

**My question here is; do you want to make a standard for all TextMate implementations, or is it okay if we take advantage of Atom's implementation?**

jeff-hykin:
To ensure this standard never has the issue of the original Textmate standard I think it should be worded as if/case statements.

For example:

```
if the token cannot be used as a variable name
    then
         it should have `keyword`
    and if all characters in the token are non-alphanumeric
        then
            it should have `keyword.symbol`
[etc]

else (if none of these cases fit)
     use the `other`
```

Even this small example exposes the ambiguity of punctuation not being able to be used as variable names, and calls into question the difference between punctuation and keywords.

jeff-hykin:
Something to consider (I haven't seen it mentioned yet in any of the PR's), is releasing all the language and theme changes as a bundled extension to get feedback. Then after some time and feedback it can be made the default, and all the old language-scopes and themes can be bundled into a "legacy scoping" extension for those who really want the old style.

It's just pretty awful when crunching to meet deadlines, and then an auto-update suddenly makes you feel blind because the colors have all changed. While on the other hand, people looking for an extension like this will be elated when they find something that does a better job at highlighting across the board.

chbk:
> My question here is; do you want to make a standard for all TextMate implementations, or is it okay if we take advantage of Atom's implementation?

I'll clarify the context of this PR and hopefully that will answer your question.

This documentation proposal is based on the Textmate naming conventions because that's what themes and grammars use, for lack of a suitable alternative. Atom defines itself as a hackable editor, yet creating a theme/grammar that's interoperable with other themes/grammars is too laborious: the Textmate scopes are insufficient and there is no Atom documentation.

Recently, Atom adopted Tree-sitter to parse source code. It works great and surpasses regex parsers in many ways: it's smarter, faster, robuster. For end users to truly appreciate this change though, we need to bridge the gap between the parser and the interface, with extensive naming conventions for grammar and theme developers.

This is where this PR comes in. Its goal is to correct and extend the Textmate conventions to make something practical for Atom. Once a clear set of conventions is endorsed and implemented, other editors that want to switch to Tree-sitter can fork the Atom grammars and benefit from them. In the long run, it's a win for everyone.

<br/><br/>

> For example consider the 'variadic sizeof' in C++ which is sizeof.... It's an operator, but it is not alphanumeric and it's also not symbolic. It's kind of both.

Taking specific examples to see how they fit in is a good way to validate the documentation. The `symbol` scope is for non-alphanumeric keywords, so it doesn't apply to `sizeof...` as a whole. However, `sizeof...` is the combination of two operators: `sizeof` and `...`. It can be split into separate tokens. The Tree-sitter grammar [does exactly that](https://github.com/tree-sitter/tree-sitter-cpp/blob/5e7476bd014445abdae879661e9caf299215478a/src/grammar.json#L5939), so each token gets its own scopes, respectively `keyword.operator` and `keyword.operator.symbol`.


<br/><br/>

> Similarly is a lambda a function definition or a variable? It's definitely both.

Scopes that are compatible with others are indicated in brackets in the documentation proposal. `entity.function` and `entity.variable` aren't considered compatible. The distinction between a function and a variable has to be clear for themes to correctly highlight them. Similarly, you could say a string literal is also a constant, but again the distinction is clear in the conventions to resolve the highlighting ambiguities that would follow.

<br/><br/>

> To ensure this standard never has the issue of the original Textmate standard I think it should be worded as if/case statements.

I'll argue that the conventions can already be read as a series of conditions: if a definition fits your token, apply the scope. There are several benefits to presenting them as a list: it avoids complicated nesting of if/else statements, authorized overlaps are explicitly indicated in brackets, it's easier to skim through to find a fitting scope, and the main scopes can be highlighted to stand out.

<br/><br/>

> Something to consider (I haven't seen it mentioned yet in any of the PR's), is releasing all the language and theme changes as a bundled extension to get feedback.

Good point, although bundling all the themes and grammars in a single package might not be feasible. I'll add some instructions in the main post for those who want to preview the PRs.

jeff-hykin:
I found Styri yesterday, and thought I recognized the profile picture. Nice job I like it 😁  👍 

So after reading your post, I think believe answer is yes (answer to "can we take advantage of Atom's implementation"). Does that sound about right?

You'll be happy to hear I'm very familiar with the Tree Sitter and TextMate! I was actually the first to get the Tree sitter working cross platform in a VS Code extension using Max's WASM build. (That extension works terribly btw). Max Brunsfield's work is amazing.

### _

> although bundling all the themes and grammars in a single package might not be feasible

Just create an empty extension that lists all the languages/themes extensions as dependencies! 👍 (assuming you've published the themes/languages as individual extensions) I learned that trick last week. I'm happy to help out in doing that. I just published a language-javascript extension last week (`this` didn't have any scopes, and JSX was missing several scopes).

###  _ 

> 1. `entity.function` and `entity.variable` aren't considered compatible. The distinction between a function and a variable has to be clear for themes to correctly highlight them.

> 2. if a definition fits your token, apply the scope.

These seem to conflict with each other, **BUT hear me out**.

Say I want all my variables **bold** and I want all my functions _blue_. 
```cpp
int i_should_be_bold = 10;
void i_should_be_blue() {};
auto i_should_be_bold_AND_blue = [ ]() -> int { return 10; };
```
I have an Atom theme that styles them correctly AND does not make a mutually-exclusive distinction. In fact I can ONLY style the third case correctly because they are allowed to overlap. I fully agree with your second statement "if a definition fits your token, apply the scope" 👍   Since a lambda meets the definition of both a variable and a function, both scopes are applied. 

Not to say that any hierarchy is bad, but any strictly-hierarchical implementation is going to be fundamentally flawed just like TextMate scopes. And Tree Sitter deserves a better standard.

###  _ 

I think you and I are on the same page in that: the standard should define what happens when there is a conflict (such as variables=red, functions=blue). In that case, I think we can both agree the `.function` class would take priority and we should document this so that theme maintainers don't have to worry about it.

#### Sizeof...

> However, sizeof... is the combination of two operators: sizeof and ...

Haha you'd think that, but actually the specification defines it as a single operator

"Variadic templates use the `sizeof...()` operator (unrelated to the older `sizeof()` operator):"

 [See here](https://en.cppreference.com/w/cpp/language/sizeof...) and [here](https://docs.microsoft.com/en-us/cpp/cpp/ellipses-and-variadic-templates?view=vs-2019)

(I maintain the C++ Textmate grammar for VS Code so I've spent and ungodly amount of time on these things). Not only is the `sizeof...()` a single operator, but the parentheses themselves are also part of the operator! (e.g. `sizeof...(Ts)` works but `sizeof...((Ts))` does not and `sizeof... Ts` also does not, you can test it [online in that first link](https://en.cppreference.com/w/cpp/language/sizeof...)) Even more strangely, despite it still being parsed as a single token/operator, gcc does allow for spaces between the `size` and `...`. 

While there is no `...` operator, there is however `...` punctuation in C++ called [parameter packing](https://en.cppreference.com/w/cpp/language/parameter_pack) there is also (a separate form of `...` punctuation) for [variadic arguments](https://en.cppreference.com/w/cpp/language/variadic_arguments). 

C++ is quite convoluted haha. The tree sitter is just making a benign mistake.

#### String literals

> Similarly, you could say a string literal is also a constant

Ruby has both immutable and mutable strings so the string itself isn't really inherently constant, even Ruby's number literals can be mutable! Javascript can actually redefine `undefined` as a function parameter so its not constant either haha (but that isn't true for `null` in Javascript). So I'd actually say that both numbers and strings are `.literal` and neither are inherently constant. Same with date literals (yaml), file-path literals (nix lang), symbols/atoms (ruby/elixir), regex, and many other edgecases that people don't think about.

Once upon a time, I thought `loop` (in Ruby) was colored incorrectly. It was being highlighted as a built-in function when it was a keyword.  ... Nope; turns out `loop` is indeed just a function, despite feeling like a keyword but I only learned that because of good syntax highlighting and theming.

Calling all literals `constant` actually can detract from people's understanding of what they are reading.

#### _

I really believe we can make a standard where there are no major/true `"you could say"`'s. Do you also think that is possible?

chbk:
> So after reading your post, I'm thinking the answer is yes (answer to "can we take advantage of Atom's [e.g. TreeSitter] implementation"). Is that okay / a correct assumption?

Yes I don't see why not let others use it.

<br/><br/>

> I was actually the first to get the Tree sitter working cross platform in a VS Code extension (using Max Brunsfield's WASM build).

That's great! I think VS Code would benefit a lot from Tree-sitter as well.

<br/><br/>

> Just create an empty extension that lists all the language & theme extensions as dependencies!

Nice! I'll try that out.

<br/><br/>

> I really believe we can make a standard where there are no major/true "well... you could say"'s. Not only possible, but easy thanks to CSS and the tree sitter. Do you think it is possible?

It's something to aim for. The conventions should settle the ambiguities so developers don't struggle with them. There's a consensus to reach between languages so developers can achieve consistent highlighting within and across them.

<br/><br/>

> Since a lambda meets the definition of both a variable and a function, both scopes are applied.

I disagree with this. I'd rather add a `lambda` scope to `entity.function`. It would be more explicit and enable the kind of customization you want. Allowing a token to be scoped both `entity.function` and `entity.variable` will just cause confusion, especially if we have to specify in what order they should appear in a CSS stylesheet for a theme to work correctly.

<br/><br/>

> Not only is the sizeof...() a single operator, but the parentheses themselves are also part of the operator!

You can write `sizeof ... ()` with whitespaces in the middle, so I believe the Tree-sitter grammar made the right choice by splitting it into separate tokens.

<br/><br/>

> While there is no ... operator, there is however ... punctuation in C++

Thanks for pointing that out, I'll change the scope to `punctuation`.

<br/><br/>

> So I'd actually say that both numbers and strings are .literal and neither are inherently constant. Same with date literals (yaml), file-path literals (nix lang), symbols/atoms (ruby/elixir), regex, and many other edgecases that people don't think about.

I agree with you, having a `literal` scope would make sense. However, in the Textmate documentation, `constant` is the conventional scope for numbers, characters, and primitive values so I kept it. Adding the `mutable` and `immutable` scopes fills the gap without needing to rename and disrupt `constant`.

<br/><br/>

> In a similar vein, regex has true conditional statements which should be scoped as control flow. However, regex anchors do not control flow (and I believe they shouldn't be scoped as control flow).

I used [this](https://github.com/atom/language-javascript/blob/16b9f3e2a846d8a4e75e5b2627bf12f642930241/grammars/regular%20expressions%20(javascript).cson#L30) as a reference for the documentation but I'm open to suggestions for better scopes.

jeff-hykin:
I think `.lambda` is a perfectly valid solution as well! It's certainly much closer to how people actually talk about it. I do want to point out though, there are still going to be a lot of weird edge cases. For example
```
const auto a_var = []() mutable {};
```
That is a global constant mutable lambda haha. In which case I still think it deserves both the constant and lambda tags.


I'm fine with things like meta, that don't really make sense but can be reused from convention. And For numbers and strings I would be fine with .constant, except that constant already has a meaning (like in my example above). π, e, MAX_INT_SIZE are regarded constants by both programmers and mathematicians, and essentially nobody ever refers to a literal as "that constant". What if you (a theme maker) want to tag all literals but not constants? You'd have to do constant:not(every-single-possible-scope-for-a-literal)


While there could be an additional ".literal" so that `:not(.literal)` works, the theme maintainer isn't going to know that immediately. And if they learn it from trial-and-error (very bad, very discouraging) they're going to think "okay literals were an exception, what else is going to be an exception? classes? containers? keywords?" They're going to have to learn the entire naming scheme before they can style something with confidence.


> I used this as a reference for the documentation

Okay I'll take a look at it and see what I can come up with 👍 

Before I plan specific suggestions though, I'm curious about the final verdict on "if it meets the definition, add it".

Specifically the code example above where `mutable` describes the function (impure function) and `constant` describes the variable (cannot be reassigned). I think there could be a simple targeted adjective system, but I'm not sure if you're interested in that.

chbk:
> except that constant already has a meaning

Different languages call similar things in different ways, [there's a lot of diversity](https://rosettacode.org/wiki/Enforced_immutability). The conventions just reach a consensus: tokens that are "constant", "final", "read-only", etc. are scoped `immutable`.


<br/><br/>

> I'm curious about the final verdict on "if it meets the definition, add it".

The conventions are a tree, so simply pick the branch whose definition fits your token. Overlaps are explicitly stated in brackets, they can be read as extensions of the branch you pick.

jeff-hykin:
Immutable, although somewhat unusual, works for me in terms of consistency (kind of like using self VS this, or property VS attribute or method VS member-function) How would you scope that function though?

```
​const auto a_var = []() mutable {};
```

`.lambda.immutable.mutable`?

chbk:
I wonder whether scoping a function as `immutable` is relevant, the `const` keyword doesn't seem to have any effect here. Also, the `mutable` keyword denotes that the function can modify immutable values, not that the function is modifiable itself. So in all simplicity, `a_var` should be `entity.function.lambda`. If you specifically want to differentiate lambdas that modify immutable values, that's outside the range of this documentation, but you are free to add an additional non-conflicting scope in your grammar, perhaps `impure` would be a fitting choice.

jeff-hykin:
> that's outside the range of this documentation

Understandable, not every edgecase should be directly covered in the standard. My intent was not solving this specific case but rather using it as an example of many situations with conflicting definitions. We can't really judge if it is "relevant" because it's not only for us. It could be extremely relevant for different people or languages. Mutability is very important in Elixir, and rust's "unsafe" could be the defining line for an banking system. Just blanket saying "this doesn't matter" is a pretty bold statement.

Language servers (not just the tree sitter) will also be using this same scoping standard, and they're going to have a lot more scopes describing variables and functions (async, safe/unsafe, pure/impure, thread-safe, signal-safe, exceptionless, capturing/non-capturing, const, constexpr, member/non-member, friend, public/private/protected, etc).

Back on topic though: for another example of overlapping definitions, consider Haskell where the operators themselves are entities that are passed around as values, yet they are also still operators. Or consider the opposite case in Haskell, where functions can be used as infix operators when inside backticks:  `1 ``plus`` 2`. 

Definitions like <br>
"all characters in the token are alphanumeric" <br>
And <br>
"At least one character in the token is not alphanumeric" <br>
are a partition.

Same with <br>
1. "all characters in the token are alphanumeric" 
2. "all characters in the token are not alphanumeric"
3. "Everything else"


My main point is; the current hierarchy isn't a partition there are gray areas exactly like the Textmate guidelines. Once grammar makers/language-servers reach those gray areas, they're going to make up their own solution. Two people make their own solution and *boom*: competing naming conventions that themers will have to memorize, grammar maintainers have to debate which of the two is "better"; and we are right back at square 1. The design of the standard should be capable of dealing with concepts for languages that haven't even been invented yet.

TLDR; Lambda is a fix for the specific case where the definitions (of variable and function) overlap, but what about all the other possible overlaps?

### Style Bleeding 

A second (more practical) issue is that the CSS classes don't support a hierarchy. For example styling `.storage.keyword` will accidentally also apply that same style to `.keyword.operator.storage` (an operator like "sizeof" that assesses the amount of storage).
Something such as `.entity.type.storage` (a definition of a memory layout on an embedded system) would accidentally get all the same styles as `.storage.type.entity`, because in CSS order doesn't really matter.

To have the scopes function as an actual hierarchy and not "bleed" style into other parts of the heirarchy, the scopes for something `storage keyword [declaration] variable` would have to be like
```
.storage
.storage--keyword
.storage--declaration
.storage--keyword--variable
```
Which is a viable solution


Non-conflicting language specific scopes could also be created such as
```
.storage
.storage--keyword
.storage--declaration
.cpp--auto
```
or 
```
.lambda
.cpp--mutable
.cpp--function
```

jeff-hykin:
Sorry, I don't mean to be so pessimistic. So far you've done a great job addressing the issues I bring up. Lots of your solutions are good answers I hadn't considered (which might be the case with the last two issues I brought up too).

I'm bringing these up because breaking backwards compatibility is going to be painful. If the change doesn't fix everything now, I people are going to be really against doing a second breaking change. If the fix doesn't clearly address all the issues, I think people will be less on board about the changes.

While I think a flat system would solve the problem, it would basically be a full re-write which isn't great. I'm open to a partitioned heirarchy. However, whichever system is picked, I don't think gray areas should be allowed.


Okay back to the details
> You can't scope something entity.type.storage because that conflicts with storage which is a higher-order scope. But you can scope something punctuation.definition.function and it won't conflict with entity.function which is another category. The predefined structure helps themes target exactly what they intend by categorizing their highlighting rules, instead of having to foresee every combination of scopes.


So you're saying the standard ensures that names don't end up in multiple places of the heirarchy? (Which would prevent the style "bleeding" problem)

I'm a bit confused with the `entity.operator` though, wouldn't that cause an overlap with `keyword.operator`? (I think there might be a point you're making that I'm not understanding).

jeff-hykin:
> But you can scope something punctuation.definition.function and it won't conflict with entity.function which is another category


Note: If someone styles `.function` it will end up conflicting by styling both punctuation.definition.function and entity.function

chbk:
Yeah ensuring backward compatibility is tough, especially when there isn't a definite standard to work with besides the scarce Textmate conventions. I'm glad you're passionate about this and I value your input and suggestions for different approaches. I think the best way to understand the hierarchy is to check out the updated [template theme stylesheet](https://github.com/chbk/apm/blob/scopes/templates/theme/styles/syntax.less). It shows how to structure a theme by separating scopes into categories.

The idea is that having a rule just for `operator` or just for `function` is too general, so you put them into categories that clarify your intent, like `keyword` or `entity`. It's already the case with most themes, you'd see a `keyword.operator` rule or `entity.function` rule. But the scopes weren't clearly organized in the Textmate documentation so I brought that all together in this PR. Another benefit is that categories also act as fallbacks. Having an `entity` styling rule provides a default for other identifiers.

jeff-hykin:
Hey, do you think we could add a prefix to these scopes to preserve backwards compatibility? 

So for example, lets say the prefix is `_`
```
   '"if"': 'keyword.control' 
```
Becomes
```
   '"if"': 'keyword.control _.keyword.control.condition' 
```
Instead of
```
   '"if"': 'keyword.control.condition'
```

If a new theme is made, it can exclusively use the standard `_.keyword.control.condition`. That way we can see when themes and grammars convert from old to new instead of us being confused about if someone is trying to use the old scopes or not.

chbk:
Nice, it's great if you can get these scopes into other grammar packages. I'm happy with anything that brings us a usable standard. Your prefix idea is feasible, however, the scopes documented in this PR really aren't that radically different from the TextMate scopes. You'll end up with a lot of `std-` repetitions of already present scopes. Also, if the intent was to write a new standard from scratch and use it alongside the TextMate scopes, I would have taken a different approach (for example I would have renamed and reorganized the `constant` scope, among other things). In short I see this PR as an extension to TextMate scopes rather than a completely new standard.

BTW the majority of scopes here aren't mandatory, you can have a very functional grammar/theme using only the bold scopes. The other scopes are here to provide plenty of examples and can be used if needed. So another possibility would be to just add some of the bold scopes to grammar packages, and then later complete them and weed out the old irrelevant scopes. This would help themes that rely on the obsolete `name` scope and the `variable` scope (some themes assume it's only used for noteworthy variables, even though there is no such restriction in the original TextMate documentation).

If you want to open a repo for this, go for it :+1:
Also be sure to check out the [Markdown file](https://github.com/chbk/flight-manual.atom.io/blob/scopes/content/hacking-atom/sections/syntax-naming-conventions.md) that's included in this PR.

PR #590, opened 2020-01-14
Update installing-atom.md

ijf8090:


lee-dohm:
Thanks for reaching out.

Since `snap` isn't an official installation method for Atom, we won't be adding a reference to using it for installing Atom in the official documentation. Thanks for the contribution though!

ijf8090:
Thanks for the feedback. If snap is not supported,  how should a user install batom on Ubuntu 18.04? 
The method in the documentation does not work.

PR #600, opened 2020-02-29
Added an online regex tester to Getting Started

cyrilbois:
### Description of the Change

I added extendsclass.com (I am the founder) to "Getting Started". It is an online regex tester with a regex visualizer which helps to understand regular expressions. 

### Release Notes

Not applicable


lee-dohm:
Thanks very much for the suggestion.

We feel that the current example sites are sufficient for beginners just getting started with regular expressions and won't be accepting this PR.

PR #606, opened 2020-03-22
Fix default keybinding for text reflow

Kleinendorst:
### Description of the Change
Fixed a wrong default keybinding mentioned in the docs.

### Release Notes
N/A


dogghousehq:
fix

PR #722, opened 2021-08-21
Update package-modifying-text.md

bsinglet:
### Description of the Change

I fixed a bug (older font name?) in the convert() function, which tries to use font  `o8` instead of `O8`. I'm not sure if this was the older name of the font, but `O8` is the current name at least as of figlet version 1.5.2. Also, updated the figlet version to the current one, 1.5.2.

### Release Notes

Updated the version of figlet and the name of the imported font.


Omphemetsepeter (approved):

icecream17 (approved):
font file at https://github.com/patorjk/figlet.js/blob/master/importable-fonts/O8.js

